PR #442, opened 2015-03-18
indoor map events & toggle location button for android

meetshawn:
I wasn't sure what branch to pull request to, since I've never done this before. Sorry if its the wrong one...

I've been using these additions for a while finally got around to putting them up. The indoor bits exactly match the original API. I would imagine you might want the location button toggle as it's own function but since its weird to access on the Java end (through getUiSettings) I just put it into setMyLocationEnabled. 

I don't use the plugin on iOS and I have no experience with objective-c, so the iOS equivalents still need to be added.

### DIFF
--- a/src/android/plugin/google/maps/GoogleMaps.java
+++ b/src/android/plugin/google/maps/GoogleMaps.java
@@ -88,12 +88,15 @@ import com.google.android.gms.maps.model.Polygon;
 import com.google.android.gms.maps.model.PolygonOptions;
 import com.google.android.gms.maps.model.Polyline;
 import com.google.android.gms.maps.model.VisibleRegion;
+import com.google.android.gms.maps.model.IndoorBuilding;
+import com.google.android.gms.maps.model.IndoorLevel;
+import com.google.android.gms.maps.GoogleMap.OnIndoorStateChangeListener;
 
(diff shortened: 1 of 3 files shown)


wf9a5m75:
Thank you for sending your pull request. I'm busy now, so let me check this later.

wf9a5m75:
Thank you for your pull request.
I'm really appreciated for your work.
I will merge your code into the `pull_442` branch after I write the code for iOS.
However I modified your code a little.

Regarding of `map.setMyLocationEnabled(true, true)`, I'd like to avoid like that.
Using two boolean parameters is uncleared what the second value is  meaning, without reading the document.

But I forgot to implement to change the value for `map.getUiSettings().setMyLocationButtonEnabled()` at the same time with `map.setMyLocationEnabled()`.
Thank you for pointing out.

In the past code, I provided the options (`map.setMyLocationEnabled()` and `map.getUiSettings().setMyLocationButtonEnabled()`) separately.
However many developers confused for that.
So I implemented them to one feature, and everyone does not argument for this.

meetshawn:
Great!

wf9a5m75:
Merge the pull request #442 is merged into the test branch.
Thank you for sending the grateful pull request, @meetshawn.
You save my time :+1:

PR #513, opened 2015-04-27
Update googlemaps-cdv-plugin.js

bairathirahul:
Converted this:
if (typeof eventName === "string" && eventName in _listeners) {

to this 
if (typeof eventName === "string") 
    	if(eventName in _listeners) {

So that if no listener is assigned to a particular event, the OFF method does not clear all event listeners when it is called for the former event.


Hirbod:
Please re-submit to #test branch

PR #517, opened 2015-04-30
The off listener method bug

bairathirahul:


Hirbod:
@wf9a5m75 the PR looks plausible to me. I think it is bulletproof for merge. You agree?

PR #529, opened 2015-05-29
Wrap marker setIcon execute in try/catch to avoid crash in host app

webprofusion-chrisc:
Markers may be invalidated and bitmaps disposed after the initial marker has been created, this can result in a fatal IllegalArgumentException. The exception can cause the host cordova app to crash therefore catching it (and warning via Log) is necessary.


Hirbod:
Very good. @wf9a5m75 I'd like to merge this one here.

Issue #526, opened 2015-05-13
TileOverlay functions not working (setVisible, remove)


canadaka:
I cannot get some of the TileOverlay function to work, .remove() does nothing. setVisible returns the following error:

"Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference"

Some sample code
````
var labels = null;
if (showLabels === true) {
	map.addTileOverlay({
	  tileUrlFormat: 'http://url/<zoom>/<x>/<y>.png',
	  zIndex: 6,
	  width: 256,
	  height: 256
	}, function(TileLayer) {
	  console.log(TileLayer.getId()); // this works
	  TileLayer.setVisible(false); // this makes an error
	  return labels = TileLayer;
	});
} else {
	labels.setVisible(false); // this makes an error
	labels.remove(); // this does nothing, but no error
	labels = null;
}
````

canadaka:
I can confirm the code is working to this point and the this.getId() returns correct.
https://github.com/wf9a5m75/phonegap-googlemaps-plugin/blob/master/www/googlemaps-cdv-plugin.js#L1682-L1685

I also added debugging logging to the plugin, and can confirm this code is firing and the "id" is correct:
id=tile_to1
https://github.com/wf9a5m75/phonegap-googlemaps-plugin/blob/master/src/android/plugin/google/maps/PluginTileOverlay.java#L85-L86

But I've found that tileOverlay == null so it never gets to the remove() function on line 91.

masashiGMS:
Maybe insert `id = id.replace("tile_", "tileProvider_");` before `TileOverlay tileOverlay = (TileOverlay)this.objects.get(id);`

canadaka:
yah I tried that, been messing with it for an hour. I finally fixed it!!
Changed line 45 to

````
this.objects.put(id, tileOverlay);
this.objects.put("tileProvider_" + id.replace("tile_", "tileProvider_"), tileProvider);
````

Hirbod:
fixed

PR #550, opened 2015-06-23
fix TileOverlay remove bug

canadaka:
tileOverlay needed to be added to the hasmap, as it is what is removde on line 86.
Tested on my phone with adb Log messages to confirm.
fixes issue #526


masashiGMS:
Did you check the `setOpacity()` method work correctly with your code?

canadaka:
I did not, I will close this request and make a new one when I confirm that function works.

PR #551, opened 2015-06-23
fix TileOverlay remove bug & tileProvider functions still work

canadaka:
tileOverlay needed to be added to the hashmap, as it is what is removed on line 86.
Tested on my phone with adb Log messages to confirm.
fixes issue #526

Also tested that setOpacity() & clearTileCache() functions work.


Hirbod:
Very good, thanks. Looks good to me. @wf9a5m75 I'd like to merge it.

Hirbod:
Thanks. Merged

PR #567, opened 2015-07-13
Remove deprecated plugin from plugin.xml; use <framework> tag

jscuba:
Google Play Services dependency was incorrectly included in Android build. It conflicted with cordova-background-geolocation and caused a build error:

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':processArmv7DebugResources'.
> Error: more than one library with package name 'com.google.android.gms'
  You can temporarily disable this error with android.enforceUniquePackageName=false
  However, this is temporary and will be enforced in 1.0

The pull request comes from discussion on cordova-background-geolocation. Credit to christocracy and lakano.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -44,7 +44,7 @@
             </feature>
         </config-file>
 
-        <dependency id="com.google.playservices" url="https://github.com/wf9a5m75/google-play-services#v23" />
+        <framework src="com.google.android.gms:play-services-location:7.3.0" />
     
         <!-- plugin src files -->
         <source-file src="src/android/plugin/google/maps/AsyncLicenseInfo.java" target-dir="src/plugin/google/maps" />


cosmith:
I have the same problem, trying to use Google Analytics, I merged your fix in my branch but it fails.
The error in the java code is:

```
Caused by: java.lang.IllegalStateException: The meta-data tag in your app's AndroidManifest.xml does not have the right value.  
Expected 7327000 but found 6171000.  
You must have the following declaration within the <application> element:    
<meta-data android:name="com.google.android.gms.version" android:value="@integer/google_play_services_version" />
```

What I don't understand is that the code is using `@integer/google_play_services_version` everywhere...

canadaka:
Hmm didn't work for me, app builds but then I get this error when trying to load a map
````
TypeError: Cannot read property 'maps' of undefined
````

I built again and this time I get
````
ReferenceError: plugin is not defined
    at getType (google-map.js:281)
````
But the `plugin.google.maps` folder is in plugins.

Issue #54, opened 2014-05-13
Possibility to pass custom variables to a marker


em-ka:
Hi,
I would appreciate a possibility to set custom variables to a marker. e.g. when you want to show additional information after user clicks on an info window (you need to pass some kind of an ID, to be able to show them).

Jerome2606:
@em-ka Could you submit a pull request for this because I need it too but I don't know how to reproduce your change ?

Thanks

PR #579, opened 2015-07-30
Pass custom variables to a marker #54

Jerome2606:
https://github.com/wf9a5m75/phonegap-googlemaps-plugin/issues/54
Add a params object to marker to add specific parameters (like Id, url, etc.) that can be use in infoClick or other events from marker.getParams()


Hirbod:
While not really necessary, I'll merge this one here

Issue #591, opened 2015-08-20
 more than one library with package name 'com.google.android.gms'


n00dl3:
Hi guys,
I'm building an application that uses this plugin and the [aerogear push plugin](https://github.com/aerogear/aerogear-cordova-push). The fact is aerogear requires `com.google.android.gms:play-services:6.1.11` whereas google maps plugin requires the version from [google-play-service](https://github.com/wf9a5m75/google-play-services) plugin. 

As the google-play-service plugin is deprecated, I tried using the recommended `<framework>` tag in `plugin.xml`, for the moment I'm trying with the 7.8.0 version, don't know if it will work with this version, but it's not important, for now I just want it to compile :-)...

So I have replaced  this:
````xml
<dependency id="com.google.playservices" url="https://github.com/wf9a5m75/google-play-services#v23" /> 
````
by this:
````xml
<framework src="com.google.android.gms:play-services-maps:7.8.0"/>
<framework src="com.google.android.gms:play-services-location:7.8.0"/>
````

However, when I run `cordova build` gradle complains :

>Error: more than one library with package name 'com.google.android.gms'

I have no idea how to handle this version conflict, any help would be appreciated....

Hirbod:
Please ask on stack-overflow

Hirbod:
Fixed in the master

PR #599, opened 2015-09-01
Update deprecated dependency for Android

denisbabineau:
dependency plugin com.google.playservices is deprecated; updating plugin.xml to use <framework> instead; pulling in the latest API

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -44,8 +44,9 @@
             </feature>
         </config-file>
 
-        <dependency id="com.google.playservices" url="https://github.com/wf9a5m75/google-play-services#v23" />
-    
+        <framework src="com.google.android.gms:play-services-maps:+" />
+        <framework src="com.google.android.gms:play-services-location:+" />
+
         <!-- plugin src files -->
         <source-file src="src/android/plugin/google/maps/AsyncLicenseInfo.java" target-dir="src/plugin/google/maps" />
         <source-file src="src/android/plugin/google/maps/AsyncKmlParser.java" target-dir="src/plugin/google/maps" />


Hirbod:
Is this tested? Does it work with a naked installation and an update? I don't have time to verify currently

denisbabineau:
Tested with:
cordova 5.0.0
cordova android platform: 4.0.0

Not sure what you mean by update? To update my plugins I rm/add, isn't this the only way?

Hirbod:
Never mind. Tested and merged. Works perfectly. Also fixed #591 
Thanks for your contribution.

nico-Fritz:
I have updated with new version of the plugin and there is already framework instead of dependency but it didn't work

denisbabineau:
can you elaborate on the error and provide cordova/platform versions that you're using? are you using cordova or phonegap?

nico-Fritz:
I'm using phonegap 5.2.0

nico-Fritz:
I have this error while I'm building android: "more than one library with package name 'com.google.android.gms' "

Hirbod:
Make sure that you have the newest Cordova / Cordova android platform installed. This is highly tested and works fine.

denisbabineau:
Tested with phonegap 5.2.0, android 4.1.1, ios 3.9.1 in a blank project and it built fine with the exception of android's minSdkVersion needing to be bumped up to 14

nico-Fritz:
ok. I resolved. Thank you!

Hirbod:
@denisbabineau, some people have weird problems after I've merged your pull-request. What is wrong? Do they need to bump the minSdkVersion? See #613

denisbabineau:
@Hirbod reading the other issues, it sounds like #613 & #606 was due to "Google Repository" missing in the SDK manager and not minSdkVersion being too low but I can't verify since I never saw those specific errors and I already had that SDK downloaded to begin with. Although it would make sense that my PR added this requirement as well by using the <framework> config to pull in the dependency.  Hopefully adding this SDK component will fix the issue for the majority of people.

The minSdkVersion seems to be related to phonegap setting a (too low?) version as an initial value for projects but I'm not sure at what version on both phonegap's and cordova's side of things these default versions got bumped. When I tested phonegap 5.2.0 in a sandbox I first got this:

```
Manifest merger failed : uses-sdk:minSdkVersion 7 cannot be smaller than version 9 declared in library /Users/dbabineau/tmp_sandbox/test/platforms/android/build/intermediates/exploded-aar/com.google.android.gms/play-services-maps/7.3.0/AndroidManifest.xml

  	Suggestion: use tools:overrideLibrary="com.google.android.gms.maps" to force usage
```

bumping to version 9 then gave me this:

```
Manifest merger failed : uses-sdk:minSdkVersion 9 cannot be smaller than version 14 declared in library /Users/dbabineau/tmp_sandbox/test/platforms/android/build/intermediates/exploded-aar/android/CordovaLib/unspecified/debug/AndroidManifest.xml

  	Suggestion: use tools:overrideLibrary="org.apache.cordova" to force usage
```

The first error made sense (play-services-maps requires 9) but the second one not so much (what's unspecified?) but in any case bumping to 14 fixed this.

I ran a few more sandboxed tests with other versions, here are the results:

- phonegap 5.3.1, cordova-android 4.1.1, android SDK 22: I got the following even before adding the plugin (just building hello world android project), I might have gotten the same error with 5.2.0 but I probably ran things in a different order:
```
Manifest merger failed : uses-sdk:minSdkVersion 7 cannot be smaller than version 14 declared in library /Users/dbabineau/tmp_sandbox2/test/platforms/android/build/intermediates/exploded-aar/android/CordovaLib/unspecified/debug/AndroidManifest.xml
```
after fixing this, plugin installed and built ok.

- cordova 4.3.1, cordova-android 3.7.2, android SDK 21: failed with the following ...
```
Failed to install 'plugin.google.maps':TypeError: Uh oh!
Path must be a string. Received undefined
```

- cordova 5.0.0, cordova-android 4.0.0, android SDK 22: OK

Sorry for the headaches, so many build environments, so little time to test! :-\

I hope this is helpful!

Hirbod:
Thanks. I've already changed the wiki with a hint to the repos.

Issue #630, opened 2015-09-17
Switch to com.synconset.cordovahttp [$150]


jperl:
[phonegap-googlemaps-plugin](https://github.com/wf9a5m75/phonegap-googlemaps-plugin) depends on [plugin.http.request](https://github.com/wf9a5m75/phonegap-googlemaps-plugin/blob/64fa35f3ba04d43557e47d614a8d0061993457ad/plugin.xml#L21) which depends on google-volley through a [jar file](https://github.com/wf9a5m75/phonegap-http-request/blob/master/plugin.xml#L21
). This makes it incompatible with other plugins that rely on `google-volley`.

[com.synconset.cordovahttp](https://github.com/wymsee/cordova-HTTP) does not rely on google volley, has no [`jar`](https://github.com/wymsee/cordova-HTTP/blob/master/plugin.xml) dependencies, and is up-to-date / better supported plugin than `phonegap-http-request`. So switching to that should fix this issue as well as allow `phonegap-http-request` to be deprecated.

-----------------------

UNEXPECTED TOP-LEVEL EXCEPTION:
com.android.dex.DexException: Multiple dex files define Lcom/android/volley/VolleyError;
at com.android.dx.merge.DexMerger.readSortableTypes(DexMerger.java:596)
at com.android.dx.merge.DexMerger.getSortedTypes(DexMerger.java:554)
at com.android.dx.merge.DexMerger.mergeClassDefs(DexMerger.java:535)
at com.android.dx.merge.DexMerger.mergeDexes(DexMerger.java:171)
at com.android.dx.merge.DexMerger.merge(DexMerger.java:189)
at com.android.dx.command.dexer.Main.mergeLibraryDexBuffers(Main.java:502)
at com.android.dx.command.dexer.Main.runMonoDex(Main.java:334)
at com.android.dx.command.dexer.Main.run(Main.java:277)
at com.android.dx.command.dexer.Main.main(Main.java:245)
at com.android.dx.command.Main.main(Main.java:106)

matsprea:
It's possible to use gradle to include and compile google-volley, but it will be via an ["unofficial" maven mirror](http://stackoverflow.com/a/26587166/3463707) (ex. [mcxiaoke/android-volley] (https://github.com/mcxiaoke/android-volley)) and it will reamain a *third party* libray so another plugin could include a different version and you will get the same problem.

[...]

PR #633, opened 2015-09-18
Switch to com.synconset.cordovahttp [#630]

matsprea:
Switch to com.synconset.cordovahttp [#630]


Hirbod:
Thanks for sharing this, but I'm not going to merge. I will drop external downloads with third party plugins and use native functions

matsprea:
@Hirbod thanks for your wonderful plugin!

PR #691, opened 2015-10-23
Fix path icon

arilsonkarmo:
In Android: Fix icon not loading when path not starting with ./
In iOS: Fix icon not loading when path start with ./


Hirbod:
@arilsonkarmo thank you.
Could you please change your Xcode settings to following and update the PR:

bildschirmfoto 2015-10-27 um 13 56 38

Hirbod:
Fixed by another pr

PR #755, opened 2015-12-04
Prevent crash on unset background color

smcpjames:
HTMLColor2RGBA() tries to read its parameter as a string. If it receives undefined, it throws an uncaught TypeError. Filter input on the front end to prevent.


Hirbod:
Thanks, looks good!

PR #761, opened 2015-12-08
Remove default addHole definition

smcpjames:
Prevents native code exception when addHole is not defined upon calling addPolygon in JS.


Hirbod:
removing a feature I've added yesterday is not PR :)

smcpjames:
This wouldn't have removed the feature, just the default empty array. addHole would still be passed as part of the polygonOptions object over to the native code which would run it as normal.

Your posted fix accomplishes the same goal, of course, but this would have also worked.

Hirbod:
Of course, you're right. It was 6am in the morning, brain was still sleeping :D

PR #764, opened 2015-12-08
Somes fixes on javascript

FmeuMan:
Fixes a color conversion issue when calling Map.setOptions(...) multiple times
Fixes InfoWindow showing on markers with title "undefined"


Hirbod:
Thank you. But there was a bug which lead into a crash when title was an integer. Thanks why I converted to string. Did you have take this into account?

FmeuMan:
Oh ok i'll take a look

FmeuMan:
Ok it's working with a number, boolean, empty string, array and object. The only quirk i found is the null value becoming "null".

Hirbod:
Well that's fine! Thanks!

Issue #638, opened 2015-09-21
CrossWalk, Google Maps Plugin and SplashScreen results in a white screen overlay


pbrisson:
Hello,

First, I know this has been discussed previously in multiple issues; I dug all reported issues and unfortunately couldn't find a solution to what I'm experiencing.

I'm using Crosswalk on Android 4.1.1. Here's my list of installed plugins:

    cordova-plugin-console 1.0.1 "Console"
    cordova-plugin-crosswalk-webview 1.3.1 "Crosswalk WebView Engine"
    cordova-plugin-device 1.0.1 "Device"
    cordova-plugin-splashscreen 2.1.0 "Splashscreen"
    cordova-plugin-whitelist 1.0.0 "Whitelist"
    plugin.google.maps 1.3.1 "phonegap-googlemaps-plugin"

As you can see, I have both splashscreen and whitelist installed, so it's unrelated to previous issues.

In order to replicate, follow instructions here: https://github.com/wf9a5m75/phonegap-googlemaps-plugin/wiki/Tutorial-for-CrossWalk-Webview-Plugin-%28Android%29

Everything works fine after compiling project after these steps.

You can replicate the issue by commenting the line:

    map = plugin.google.maps.Map.getMap(div);

Doing so will result of a white screen staying over the webview after splashscreen is hidden. There's no way to hide it. My assumption was that it could be CrossWalk splashscreen not getting first paint event so that it's not getting hidden, but I'm not sure.

I tried a bunch of things to solve this, but has been unsuccessful so far.

My issue is that I'm having a home screen on my app that allows the user to select various options before the map is shown, so there's no need to initialize the map plugin until these options has been set. Doing so result is this white screen overlay; it seems the google maps plugin absolutely has to initialized on the first screen for this to be gone.

I'm callling it an overlay because the white screen is only hiding the webview; it's still getting clicks! So if I get the overlay, then clicks on the area where a button is supposed to be, it gets me to the second screen where the map is initialized and the white screen disappears.

Please note if I launch the app on my phone (Samsung Galaxy S5), and that the lock screen is still active when I'm launching the app, I'm not a able to replicate; I need to see the splashscreen of the app while it is being launched to be able to get this white screen.

Any ideas?

Thanks for your time!

Hirbod:
Just as a quick workaround: initialize it first and hide / remove the map afterwards. Not really nice, but would this fix you problem?

[...]

PR #778, opened 2015-12-14
Fixes blank screen issue with crosswalk

FmeuMan:
For issue #638. The white screen is due to setting a white color on the parent container of the webwiew when the plugin start. I think this container is only refreshed when the map is added on background. So if no map is added on the first screen the container stay white.

What i tried :
- map creation at startup (on deviceready) then hide splashscreen on map ready or even later
- no map at startup and getMap(...). Only call it later manualy
- reduce the app and reopen id
- close the app with back button and reopen it
- kill the app and reopen it
- navigator.app.exitApp() and reopen it

What i have not tried : ideas are welcome ;)

Tested with and without crosswalk using cordova 5.3.3 and 5.4.1 with these pugins
```
cordova-plugin-crosswalk-webview 1.4.0 "Crosswalk WebView Engine"
cordova-plugin-device 1.1.0 "Device"
cordova-plugin-splashscreen 3.0.0 "Splashscreen"
cordova-plugin-whitelist 1.2.0 "Whitelist"
plugin.google.maps 1.3.8 "phonegap-googlemaps-plugin"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Please consider to make a small donation. We spent many hours to make this a gre
 
 Before I release the clustering feature, I want to raise a funding of 1000$. I will donate some of the funded money to @limeuwe for donating the code and to my developer for the Android part.
 
-**Currently funded**: 50 $
+**Currently funded**: 65 $
 
 ###Quick install
 


Hirbod:
Thank you very much. I will evaluate your fixed later and when it works as intended, I will merge your code

Hirbod:
@FmeuMan do you have some time to also include your fixes into https://github.com/mapsplugin/cordova-plugin-googlemaps/tree/clusterer ?

The Codebase differes a bit, as there is another controller-setup. You would help me much with that.

FmeuMan:
I'll have a look

FmeuMan:
A more simple version from this PR is #793.
I'll do the custerer branch in the row.

PR #827, opened 2016-01-30
Fix typo in HTML_COLORS

xming13:


Hirbod:
Wow. Thanks. I don't know where Masashi got the list :D / it is full of errors. Could you also check if the Readme is fine and fix it there too?

xming13:
The list is correctly just that some of the 'd' is changed to 'c'. I have updated https://github.com/mapsplugin/cordova-plugin-googlemaps/wiki/Available-HTML-colors

Issue #830, opened 2016-01-30
map.clear() does not free memory


pbudner:
Hello everbody,

it seems that map.clear() does not sufficiently clear the memory.

The XCode memory report shows me the following statistics before adding the map and the markers:
<img width="416" alt="step0_beforemapisloaded" src="https://cloud.githubusercontent.com/assets/7352912/12697677/f7c3cfa0-c789-11e5-874e-80bf3dafb96f.png">

Adding the map and about 100 markers it looks like that:
<img width="428" alt="step1_addingalotofmarkers" src="https://cloud.githubusercontent.com/assets/7352912/12697687/0a992e2c-c78a-11e5-9b04-1d1b871fdc65.png">

Even after clearing the map via map.clear() the memory use stays the same. Does anybody experience the same issue? I am using the newest version v.1.3.9.

Removing each marker via marker.remove() does not work either. The markers are not visible on the map and can not be accessed but the memory use does not drop.

Hirbod:
Well I guess that Google itself does not use a lot of memory. It's more about the technique behind that. The plugin makes a lot of elements transparent, including the webview itself. I guess most benefit would be to call .remove() when map is not needed. Further, you should also look for ._gmaps_cdv_ and remove the background attribute after you've removed the map

pbudner:
Initializing the plugin and adding just a few markers results in the following memory utilization:
<img width="415" alt="step1_mapandmarkersloaded" src="https://cloud.githubusercontent.com/assets/7352912/12697802/4fb5711c-c78c-11e5-8cb2-8d1c8666c664.png">

So its definitely caused by adding the custom-icon markers.

Hirbod:
Do you resize them ?

pbudner:
Nope. Im just doing the following:
```
mapRef.addMarker({
       "position": markerPosition,
       'icon': markerIcon
}, function (marker) {
        mapRef.on("someEvent", function(eventArgs) {
               // do something with this event
        });
        marker.addEventListener(plugin.google.maps.event.MARKER_CLICK, function () {
               // show infos
        });
});
```

wf9a5m75:
Does this issue still problem? The pull request #832 has been already merged.

Hirbod:
ping @ all - does the problem still exists on the master branch?

sam2x:
[...]

PR #832, opened 2016-01-31
Fix clear function to clear MARKERS and KML_LAYERS too

xming13:
Might be related #830


Hirbod:
Thanks. Looks good. Will merge after testing

PR #861, opened 2016-02-24
fix #856: Bad image quality with toDataURL()

### LINKED ISSUE #856: Bad image quality with toDataURL()
We are using `toDataURL()` when the map should move similar to #50. However, the generated image is of bad quality. For testing, we removed the following lines in the `toDataURL` method in PluginMap.java:

    float density = Resources.getSystem().getDisplayMetrics().density;
    image = PluginUtil.resizeBitmap(image,
                                    (int)(image.getWidth() / density),
                                    (int)(image.getHeight() / density));

After that change, the generated images looks exactly like it looked on the native map. Is there any way to disable or properly configure that resize without changing the plugin's code?

troxler:


wf9a5m75:
This is adapted part of #856. Thanks.

PR #864, opened 2016-02-25
Fixing bug that add ./ and then the url to the icon value when is passed a base64 image

developez:
I added also some debug information under (isDebug variable).

The key is mainly add (iconUrl.startsWith("data:image") == false) to line 695. If you don't put this, a icon value with "data/image ..." is modified by the adding of "./".

I don't know why I am the only one with this problem.


Hirbod:
Please clean up you project, only having one commit, not 19

developez:
The rest is for debug information, all under isDebug, but ok I understand that is not secure, I clean the code, wait a minute please.

Hirbod:
And makes it hard to read the changes

developez:
@hirbod I reset all the code and now in my fork there is only that change, how can I reverse this pull request, the list of commit is still there.

PR #866, opened 2016-02-25
Fixing bug that add ./ and then the url to the icon value when is passed a base64 image

developez:
Adding `iconUrl.startsWith("data:image") == false` to line 682 you avoid that the string "./" was added to the icon content when this is a base64 image.

I don't know why I am the only one with this problem.

(Note: I deleted and created again the repository (muerto el perro se acabo la rabia)).


wf9a5m75:
@developez Thank you for fixing.

bentomas:
I think we need this same fix here for iOS:

https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/master/src/ios/GoogleMaps/Marker.m#L608-L616

at least I needed to add the following to get it working locally...

```
range = [iconPath rangeOfString:@"data:image"];
if (range.location != 0) {
```

PR #916, opened 2016-03-29
Rename package/plugin to prevent remapping

eg271:


Hirbod:
Just started a discussion about that with @wf9a5m75 - let us evaluate everything - then we will merge your request. Thanks so far!

Hirbod:
@eg271 could you also change the line here?
https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/master/package.json#L6

it's easier to have that in one commit :)

eg271:
@Hirbod not sure I follow - would you like me to merge both commits into 1? Is there an easy way to do that just using Github?

Hirbod:
That's possible but not necessariy for 2 changes. Thanks

Issue #934, opened 2016-04-12
Use local filesystem images for TileOverlay


webalchemist:
The plugin used to be able to handle local images for TileOverlays. When trying to upgrade my app I couldn't get this working again.

I'm not sure when the change was introduced, but PluginTileProvider.java now explicitly limits the tileUrlFormat to starting with "http://" and "https://". URLs starting with "file://" will never work.

Its fairly straight forward to get this working again for Android, but I don't know enough to do the iOS coding.

This would be really useful for caching map tiles for use when a device is not online.

wf9a5m75:
Android or iOS?

wf9a5m75:
It seems I did not implement `file://` for Android

jlsuarezs:
Any way to do it using Local Tiles? Thank you!

webalchemist:
I've put in a pull request to fix this on Android (it already works on iOS).  Hopefully they'll merge it.
If you need it working before that, you can see my suggested changes at my fork of this repo at
https://github.com/webalchemist/phonegap-googlemaps-plugin

This version is working fine for me.

wf9a5m75:
Thank you for long-time waiting.

The v2-beta version supports local tiles both Android and iOS.

```js
map.addTileOverlay({
    // Load image files from the local file path
    tileUrlFormat: "../images/map-for-free/<zoom>_<x>-<y>.gif"
}, function(tileOverlay) {

});
```
https://github.com/mapsplugin/v2.0-demo/blob/master/www/TileOverlay/addTileOverlay.html#L42-L47

The v2-beta is available.
https://github.com/mapsplugin/v2.0-demo

PR #935, opened 2016-04-12
allow local images to be used for TileOverlays

webalchemist:
Issue #934
I haven't tried building on iOS yet - so it may already work there. I didn't see anything in the iOS code that restricted the url type.

Here a pull request for my changes to the Android code.


wf9a5m75:
Thanks. I'll check this and iOS later.

wf9a5m75:
The next version allows to use local images for TileOverlay class.
I don't merge your pull request into the current master, but thank you anyway.

PR #937, opened 2016-04-13
fix a bug who send wrong list of DOM map children

dagatsoin:
I am using some functions in a Mapbox port and I found a bug when treating the children element of a map in `getAllChildren()`

For exemple:
```
<div id="map_canvas">
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
 <button>
<div>
```
will send near 100 children instead of 14. It is because the search function is launched for each root children instead of the first one. So that duplicate the list.

### DIFF
--- a/www/googlemaps-cdv-plugin.js
+++ b/www/googlemaps-cdv-plugin.js
@@ -2697,9 +2697,7 @@ function getAllChildren(root) {
             node = node.nextSibling;
         }
     };
-    for (var i = 0; i < root.childNodes.length; i++) {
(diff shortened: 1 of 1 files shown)


wf9a5m75:
In the v2-beta, the plugin uses different approach.
The plugin manages all html elements of the html.
So you don't need to face this error again.
Try the new version.
https://github.com/mapsplugin/v2.0-demo/


Thank you for sending your pull request anyway.

Issue #972, opened 2016-05-18
How to hide directions button


R2D221:
Hi, I'm starting to use this plugin, and so far it works great. However, I've found something that apparently isn't documented anywhere.

When I tap on a marker, in the bottom right corner two buttons appear, one for opening the location in Google Maps, and the other one for getting directions (using Google Maps as well). Since I don't need these functions for the scope of my app, it would seem weird to leave them there.

Is there a way to hide them using the API, or is this function not implemented yet?

![screenshot_2016-05-18-14-07-47](https://cloud.githubusercontent.com/assets/774575/15372667/26d11b0e-1d07-11e6-8ee7-62b64861eabf.png)

masashiGMS:
It comes from the Android API, not this plugin.

R2D221:
I mean, yes, but I can also hide the current location using `setMyLocationEnabled(false)`, which is ultimately a function from the native plugin.

roblav96:
This is a must, if possible please :) 

I'm going to try and switch my project from mapbox-gl-js to this plugin now that it supports ios wkwebview.

smcpjames:
Since this function is easily enabled/disabled through the Google Maps Android API's UI settings, I went ahead and added it. Please see PR #983.

roblav96:
@smcpjames I wub you =]

Slavrix:
+1 for merge

smcpjames:
@roblav96 @Slavrix If Masashi decides not to merge, you can try it from my repo: https://github.com/smcpjames/cordova-plugin-googlemaps/tree/master Just follow the install instructions under 'github' and substitute 'phonegap-googlemaps-plugin' for 'smcpjames'.

wf9a5m75:
https://github.com/mapsplugin/cordova-plugin-googlemaps/pull/983#issuecomment-223649572

wf9a5m75:
Please use the v2-beta version.
In the v2-beta version, many bugs are fixed and some new options are implemented.
This request is also implemented.

https://github.com/mapsplugin/v2.0-demo/blob/master/www/Map/setOptions.html#L51

Read the description for the v2-beta.
https://github.com/mapsplugin/v2.0-demo/

PR #983, opened 2016-05-30
Add setMapToolbarEnabled function for Android

FPCSJames:
This adds a setMapToolbarEnabled function, and a mapToolbar option during map initialization, to hide or show the [map toolbar](https://developers.google.com/maps/documentation/android-api/controls#map_toolbar) on Android devices. It also adds a stub so that calling the function on an iOS device won't result in an error.

Addresses #972.

### DIFF
--- a/src/android/plugin/google/maps/PluginMap.java
+++ b/src/android/plugin/google/maps/PluginMap.java
@@ -38,12 +38,12 @@ public class PluginMap extends MyPlugin {
   /**
    * @param args
    * @param callbackContext
-   * @throws JSONException 
+   * @throws JSONException
    */
   @SuppressWarnings("unused")
   private void setOptions(JSONArray args, CallbackContext callbackContext) throws JSONException {
 
-    
+
     UiSettings settings = this.map.getUiSettings();
     JSONObject params = args.getJSONObject(1);
     //controls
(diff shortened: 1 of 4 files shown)


wf9a5m75:
I want to implement all features for both OS: Android and iOS.
So if you want to merge this, please create the toolbar for iOS.

smcpjames:
While I understand your goal, it seems a bit nonsensical to try to implement, from scratch, a feature that Google themselves only put into Android, only so that the people who didn't want to see it to begin with can disable it?

wf9a5m75:
>There's support for the iOS-only zIndex property of Marker, the Android-only zoom option in the constructor and setOptions() of Map, and probably more.

These properties are introduced by Hirbod while I left this project for a while.
He broken the rule.

wf9a5m75:
As I wrote above, just create a toolbar for iOS, even if you use it or not.
Otherwise, please wait until when I have enough time.
My private time is really limited (one hour per day, or less), and I'm focusing on implementing multiple maps (working on the multiple_maps branch).
The code will be changed drastically on next version.
(That's why I've been stopping recently merging.)

wf9a5m75:
I implemented to the v2-beta for this feature.
But thank you for sending your pull request.
https://github.com/mapsplugin/v2.0-demo/blob/master/www/Map/setOptions.html#L51

FPCSJames:
Hope it proved useful.

Issue #1023, opened 2016-07-08
Android fromPointToLatLng function on master not called


Hurinj:
Hello,
I used the fromPointToLatLng function in 1.3.9 plugin but i had to go to 1.4.0 master because there was an error permission ACCES FINE LOCATION and ACCESS COARSE LOCATION.
But now it seems the master doesn't recognize this function, i put an alert on it and nothing appears.

Is it because this function is not yet ready in master ?
How can i delete this error permission in 1.3.9 ?

```
map.fromPointToLatLng([a , b], function (point) {
		                console.log('mousedown lat' + point[0] + ' lng:' + point[1]);
                                    map.addMarker({
                                    'position': {
                                        lat: point[0], 
                                        lng: point[1]
                                    },
                                    animation: plugin.google.maps.Animation.BOUNCE,
                                    zIndex: 100
                                    }, 
                                });
        });
```

wf9a5m75:
I guess you face another problem.
The fromPointToLatLng() method is still implemented the same as before.
(I don't change it at all)
https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/master/www/googlemaps-cdv-plugin.js#L965

Check your cordova browser with the chrome inspector. (`chrome://inspect`)

Hurinj:
The fact is nothing appears even in the inspector. Here a and b are in the console but not the "ok" and the marker doesn't appear. It's like the function is invisible..

console.log(a);
console.log(b);
map.fromPointToLatLng([a , b], function (point) {
        console.log("ok");
        });

But indeed, this function works great on my iOS 1.3.9.
I tried to install the master on iOS but i didn't succeed, do you have the right command line ?

celron:
On android, I believe that this change 74c501a67a7b9ab112ecd62f55d8039bb31c2312 or patch 1012 has a slight bug,
In PluginMap.java lines (594-595)
```
    pointX = args.getDouble(0);
    pointY = args.getDouble(1);
```
Should be
```
    pointX = args.getDouble(1);
    pointY = args.getDouble(2);
```

As I see error messages in DDMS regarding incorrect parameters.

08-04 11:22:34.211: W/System.err(15840): Caused by: org.json.JSONException: Value Map.fromPointToLatLng at 0 of type java.lang.String cannot be converted to double
08-04 11:22:34.211: W/System.err(15840): 	at org.json.JSON.typeMismatch(JSON.java:100)
[...]

PR #1065, opened 2016-08-12
Fixed json parameters for PointToLatLng

celron:
This fix the parameters that are used in the function fromPointToLatLng() so the callback will be executed.  Otherwise an error message in the android logs will appear with

08-04 11:22:34.211: W/System.err(15840): Caused by: org.json.JSONException: Value Map.fromPointToLatLng at 0 of type java.lang.String cannot be converted to double
08-04 11:22:34.211: W/System.err(15840): at org.json.JSON.typeMismatch(JSON.java:100)
08-04 11:22:34.211: W/System.err(15840): at org.json.JSONArray.getDouble(JSONArray.java:353)
08-04 11:22:34.211: W/System.err(15840): at plugin.google.maps.PluginMap.fromPointToLatLng(PluginMap.java:594)

This has been reported in issue #1023


wf9a5m75:
Thanks.

PR #1077, opened 2016-08-22
Call postExecute on callback to setIcon_ from AsyncTask.postExecute()

chris-armstrong:
This patch moves some of the error handling from the doBackground() part of the AsyncTask in PluginMarker._setIcon() to the postExecute() part. This ensures that later Google maps interaction is on the UI thread instead of a background thread (which is not allowed and causes the application to crash).


wf9a5m75:
Thanks

PR #1104, opened 2016-09-16
Bugfix iOS portrait display when in landscape mode

paulklinkenberg:
When my iPad, iOS 9.3.5, is in landscape mode, and I open the map, the map will show top-left, but only 768 px. wide, and without the footer and the Close button visible.
This also happens when rotating the device while the map is already visible.

### DIFF
--- a/src/ios/GoogleMaps/GoogleMaps.m
+++ b/src/ios/GoogleMaps/GoogleMaps.m
@@ -532,10 +532,13 @@
     // iOS8 or above
     direction = [UIDevice currentDevice].orientation;
 #endif
-
-
-    if (direction == UIInterfaceOrientationLandscapeLeft ||
-        direction == UIInterfaceOrientationLandscapeRight) {
+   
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Could you provide the test code (simple HTML that reproduces your issue)?
And which iPad do you use?

paulklinkenberg:
Hi, the issue occurs with every code example given in this repo, when the app is started in landscape mode. I created a working test cordova project here: https://github.com/paulklinkenberg/googlemaps-cordova-testproject
My iPad type is: Apple iPad Air 2 WiFi 64GB Grey
Running on iOS 9.3.5

Image example of how the map displays: https://www.icloud.com/sharedalbum/en-us/#B02532ODWBUW2n;E8A47690-8E5E-414B-8C56-08C1CA577286 (the grey part on the right is the background-color of the index.html page)

wf9a5m75:
Confirmed. Thank you for your work, and giving a test project.
Providing a pull request with test project and conditions(your environment) reduce my work a lot!

PR #1108, opened 2016-09-21
add NSLocationWhenInUseUsageDescription to parent

Fantomax:
There was issue with cordova build command. That occured because of empty parent attribute of NSLocationWhenInUseUsageDescription


wf9a5m75:
Thanks

PR #1110, opened 2016-09-23
fix contains functionality

Fantomax:
Method "contains" doesn't work correctly in situation when -180:180 lng line is in visible region. I've added another response in this case.


wf9a5m75:
Thank you. Could you show me some small test code?

wf9a5m75:
thanks

PR #1115, opened 2016-09-26
Fixed icon anchor moved in setInfoWindowAnchor

jbergasa:
setInfoWindowAnchor called setGroundAnchor, which causes some marker icons to be incorrectly anchored to the same anchor as the InfoWindow and "move around" as the user zooms in and out.

This call seems to be a copy & paste error and has now been removed.


wf9a5m75:
Thanks.

jbergasa:
Thank you for the plugin! It's incredibly useful.

Issue #1153, opened 2016-10-25
Crash if icon image is not found


samedii:
App crashes due to FileNotFoundException and IllegalStateException if icon image is not found

wf9a5m75:
By the way, the master branch (v1.3.9) will be deprecated in near the feature.
I recommend you use the v2-beta version.
https://github.com/mapsplugin/cordova-plugin-googlemaps/wiki/v2-beta

PR #1154, opened 2016-10-25
Avoid crashing if icon image is not found #1153

samedii:
App crashes due to FileNotFoundException and IllegalStateException if icon image is not found


wf9a5m75:
Thanks!

Issue #14, opened 2014-04-17
Google map Librarie


TolotraRam:
Hi,
Nice work, but i'd like to know how to implement Gmap Libraries such as PlaceService, should i use javascript or you need to implement it in your plugin ?

Hirbod:
I guess, for now, Places over JS API V3 should be enough. Places support could be an own plugin and should not be included by default. (just prevent overhead as much as possible)

wf9a5m75:
I'm waiting Google Play Services v23 which is support Places API.

wf9a5m75:
@Hirbod 
If you have a time, would you like to try to implement [Places API](https://developers.google.com/places/) or [Directions API](https://developers.google.com/maps/documentation/directions/?hl=en) for iOS using http-request plugin?

I think the directions API is easy.
If you are interested in, let's make a fund-raising for you.

Jerome2606:
Are you planning to work on Places API ?

Hirbod:
Currently not.

dflourusso:
Hi, are there some prevision from directions implementation? Thanks.

PR #1177, opened 2016-11-07
Fix always zoom in animateCamera and moveCamera #14

samedii:
Alternative fix might be to change how we convert the argument in Map.js, this line is the underlying issue:
cameraPosition.target = common.convertToPositionArray(cameraPosition.target);

I didn't want to change Map.js because that could break something in iOS. I feel like this is a design choice of the plugin that you @wf9a5m75 are more suited to make. Feel free to throw away this PR and change Map.js if you feel that's a cleaner solution


wf9a5m75:
Thank you for sending a pull request. I will check it tonight.

wf9a5m75:
Thanks. I fixed this issue in the Map.js. But your inspection saved my time.
https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/8b5fe7386e2a383bab6bbc5f94d13bb1e0308408

PR #1196, opened 2016-11-21
Add ‘tappable’ option to overlays

bentomas:
Uses native "tappable" on iOS, and uses custom one on Android by adding a "tappables" hashmap to `MyPlugin` and keeps track of which overlays should be added.


wf9a5m75:
Thank you for sending a pull request. But it's already implemented in the v2(next version).
I don't develop for the v1 anymore (except fixing critical issue)

wf9a5m75:
Since this is already implemented in v2, I close this pull request. But thank you for sending a pull request anyway :)

PR #1212, opened 2016-12-06
Unpin play services versions from 9.8.0 to +

zwacky:
should address issue #1208.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -45,8 +45,8 @@
             </feature>
         </config-file>
 
-        <framework src="com.google.android.gms:play-services-maps:9.8.0" />
-        <framework src="com.google.android.gms:play-services-location:9.8.0" />
+        <framework src="com.google.android.gms:play-services-maps:+" />
+        <framework src="com.google.android.gms:play-services-location:+" />
 
         <!-- plugin src files -->
         <source-file src="src/android/plugin/google/maps/AsyncLicenseInfo.java" target-dir="src/plugin/google/maps" />


wf9a5m75:
http://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

Fantomax:
@wf9a5m75 Could you please change it to 10.0.0? This version works good for me

djett41:
@zwacky I believe this plugin uses code that was only available a of 9.8.  So just setting to + will mean that the CLI would allow you to build with verison 8.X which would then cause class not found errors.  So this just ensures youre building for a compatible version.  If he changed 9.8 back to +, then people would file bugs about this not building on previous versions.

@wf9a5m75 changing to 10.0.0 is the opposite of what you would want to do because then youre enforcing that people need to be building with google play version 10.0.0 when it still works with 9.8

So say for instance you use another plugin that only works in 9.8, and then you need to use this plugin, your app wouldnt work because you would have google play services version mismatches..

The concept is similar to Maven/Gradle in which your project/jar/plugin specifies dependencies, and then you can override if need be.  So in this case, the best thing to do would be to have a build-extras.gradle file and override the version to match your version.. for instance

```
ext.playServicesVersion = '10.0.1'

ext.postBuildExtras = {

    configurations.all {
        resolutionStrategy {
            force "com.google.android.gms:play-services-location:$playServicesVersion"
            force "com.google.android.gms:play-services-gcm:$playServicesVersion"
            force "com.google.android.gms:play-services-maps:$playServicesVersion"
            .... any other version overrides
        }
    }
}

```

Marcus-L:
The pinned version caused me problems as well. I'd be in favor of unpinning.

PR #1245, opened 2016-12-27
Fix marker visible option

katsuhidem:
Can't add hide marker


wf9a5m75:
Since this is already implemented in v2, I close this pull request. But thank you for sending a pull request anyway :)

PR #1269, opened 2017-01-18
Corrected parameters in call to removeObjectForKey

thyde1:
The keys in mapCtrls are mapIds, not mapCtrl objects.

This fix means that when iOS maps are removed, they are actually removed, as before, they were not.

This caused issues when placing a map in a div which previously contained a map which had since been removed. This was because the plugin would pass any touch events on the new map to the old (removed) map.


NoNameProvided:
This code entirely missing from the current release, I see this PR targets the `mapsplugin:multiple_maps` can I just grab this one instead of master? Is it a drop in replacement for the current active release?

thyde1:
@NoNameProvided This PR is only a fix for the multiple_maps branch. MyPluginLayer.m appears to have changed too much between the master branch and multiple_maps for this fix to be of any use in the master branch.

Issue #1295, opened 2017-02-08
Pressing marker slides a navigation and a google maps button on to the map


JeffreyJamesMMM:
I had to re-add my plugin today and now I'm getting new behavior and I hope that there is some map or marker option to disable it. It happens when a marker is pressed. It hides if I press anywhere else on the map.  It might be undocumented because I'm not seeing it anywhere. 

![image](https://cloud.githubusercontent.com/assets/942571/22748430/0e885c84-edf8-11e6-9783-bfe9a265b77d.png)

wf9a5m75:
If you use the plugin v2, there is the `mapToolbar` option to disable it. No option for the plugin v1.
https://github.com/mapsplugin/v2.0-demo/blob/15be5b34f87b465f6b4420c7eb18d49d370dcee5/www/Map/setOptions.html#L51

wf9a5m75:
The tool bar comes from the native api.

JeffreyJamesMMM:
Thank you so much for your assistance.

wf9a5m75:
:)

PR #1300, opened 2017-02-12
feature(map settings): hide map toolbar

lpikora:
Adding option for hide toolbar when marker is clicked. Solves issue https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/1295 for plugin v 1.4.0.

Plugin v2 already supports this, but unfortunately without support of crosswalk...

### DIFF
--- a/src/android/plugin/google/maps/GoogleMaps.java
+++ b/src/android/plugin/google/maps/GoogleMaps.java
@@ -481,6 +481,9 @@ public class GoogleMaps extends CordovaPlugin implements View.OnClickListener, O
       if (controls.has("zoom")) {
         options.zoomControlsEnabled(controls.getBoolean("zoom"));
       }
+      if (controls.has("mapToolbar")) {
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Thank you for sending a pull request.
But as we already announced v1.x has been stopped. 
Please make a pull request that support crosswork with the plugin v2.

wf9a5m75:
This is already implemented as the `mapToolbar` property.
You can specify the Map.getMap() and map.getOptions().

```js
    map.setOptions({
      'mapType': plugin.google.maps.MapTypeId.HYBRID,
      'controls': {
        'compass': true,
        'myLocationButton': true,
        'indoorPicker': true,
        'zoom': true,
        'mapToolbar': true   // <---    currently Android only
      },
     ...
  });
```

And also the v2 works with cross work.
https://github.com/mapsplugin/cordova-plugin-googlemaps-doc/blob/master/v2.0.0/Installation/crosswalk/README.md

PR #1351, opened 2017-03-11
getTitle for polyline and polygon

sudhirbitsgoa:
Add method getTitle for polyline and polygon.

### DIFF
--- a/src/android/plugin/google/maps/AsyncKmlParser.java
+++ b/src/android/plugin/google/maps/AsyncKmlParser.java
@@ -245,7 +245,12 @@ public class AsyncKmlParser extends AsyncTask<String, Void, Bundle> {
           //-----------------
           // Polyline
           //-----------------
-          
+          // title
+          tmp = childNode.getString("name");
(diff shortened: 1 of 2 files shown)


wf9a5m75:
Please send a pull request for both Android and iOS.

sudhirbitsgoa:
I can't test iOS

wf9a5m75:
@sudhirbitsgoa Finally the kmlLayer is ready to go for Android and iOS.

In order to use the `map.addKmlOverlay()`, you need to install the plugin from the `multiple_maps` branch. The new version passes all attribute values such as `title` or `name`.

https://github.com/mapsplugin/v2.0-demo/blob/master/www/KmlOverlay/addKmlOverlay.html

At this time, `kmz` file is not supported. And even one kml file, not all kml tags are supported, such as `<screenoverlay>`.
I will update which tags are available later, but try the demo first.

------

Since this PR is too old, so I can't merge your code. But I appreciate for your PR.

Issue #1345, opened 2017-03-07
google map plugin isavailable  is undefined


minamorsali:
hello I want to use google map plugin with cordova 6.5.0. I before use it in cordova 6.0.0 and it is ok. but now in cordova 6.5.0 plugin.google.maps.Map.isAvailable , isavailable variable is undefined...
why and how can I use it? please help me... thank you

wf9a5m75:
I have no idea

minamorsali:
no one has this issue?

Assilfira:
If you are using V2, the plugin do that verification itself, so you don't have to do it yourself. If you are using V1, I don't know what can be the problem.

thohng:
This one works with us:
https://github.com/mapsplugin/cordova-plugin-googlemaps/pull/1398

PR #1398, opened 2017-04-12
fix cannot read property 'maps' of undefined on android-19

thohng:
workaround with issue `cannot read property 'maps' of undefined` on android-19 or without AOT compilation


thohng:
#1345

wf9a5m75:
Thank you for your help

PR #1412, opened 2017-04-21
Poly edition

ntorrescodigodelsur:


wf9a5m75:
Thank you for sending pull request. It seems wonderful feature.
However, the version 1.x will be shutdown in this year (not decided particular date).
https://github.com/mapsplugin/cordova-plugin-googlemaps#version-20-beta-roadmap

And also, using the plugin v2, it's easy for Android and iOS
https://github.com/mapsplugin/v2.0-demo/blob/master/www/Polyline/getPoints.html

PR #1418, opened 2017-04-25
Fix getCameraPosition()

kostap13:
getCameraPosition() not fire callback


wf9a5m75:
Thank you for sending a pull request, but actually, taking callback is wrong.

The v2 is designed the most of all methods work as `synchronously`.

https://github.com/mapsplugin/v2.0-demo/blob/master/www/Map/getCameraPosition.html#L37

PR #1422, opened 2017-04-28
Add updatePoints callback

kvilon:
It adds `updatePoints` callback for the polygon and polyline `setEditable` methods.
This callback has to be called every time user changes the polygon/polyline, **before the `overlay_edit` event is emitted**.


wf9a5m75:
Thank you for huge pull request, but what is this? Could you give me more description and demos?

And you may know what, the plugin project focus on the next version (2.0-beta).
It is significantly different from v1.

Is it possible to port your code to the v2?

PR #1425, opened 2017-04-29
Solving the memory leak

sam2x:
Removing the memory leak by explicitly setting null to some persistent references. 
Covered following method: 
- Map.destroy()

Not covered:
- Map.clear()
- Marker.destroy()

So to be explicit:
- With this patch, If the map is destroyed, all the allocated memory is fully released.
- However, all others elements destruction/clear method (eg: a marker removal) will leak memory (at least until the map is destroyed). To cover/resolve this, you need to apply the same "kind" of patch i did on map.destroy() to others elements.

### DIFF
--- a/src/android/plugin/google/maps/PluginMarker.java
+++ b/src/android/plugin/google/maps/PluginMarker.java
@@ -950,7 +950,7 @@ public class PluginMarker extends MyPlugin implements MyPluginInterface  {
             try {
               inputStream = assetManager.open(iconUrl);
               image = BitmapFactory.decodeStream(inputStream);
-              inputStream.close()
+              inputStream.close();
             } catch (IOException e) {
               e.printStackTrace();
               return null;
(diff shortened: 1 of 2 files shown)


sam2x:
Tested the PR on my app, working as expected :)

src/android/plugin/google/maps/PluginMarker.java
```diff
@@ -958,8 +959,6 @@ protected Bitmap doInBackground(Void... params) {
           if (image == null) {
             return null;
           }
-          icons.add(image);
```
wf9a5m75: Deleting this line, you are misunderstand.
Caching local image reference to the icon property is for releasing them on clear method.

```java
    Bitmap[] cachedBitmaps = icons.toArray(new Bitmap[icons.size()]);
    Bitmap image;
    for (int i = 0; i < cachedBitmaps.length; i++) {
      image = icons.remove(0);
      if (image != null && !image.isRecycled()) {
        image.recycle();
      }
      image = null;
    }
```

Deleting this line means local icon images are not released.

PR #1442, opened 2017-05-11
ARGB - fix for alpha channel

vosecek:
Alpha channel is first two signs in 8 signs colour hexa code. Used in Google Maps KML export for example.

It would be better to keep compatibility with #AARRGGBB coding style.


wf9a5m75:
The multiple_maps does not support the KML yet.

And the method name `HTMLColor2RGBA` means `RGBA` not `ARGB`

Issue #1493, opened 2017-06-13
 TypeError: node.className.indexOf is not a function. 


marijnroukens:
After updating to the latest version I receive this error message. The app is still working but the map is not loading:

` TypeError: node.className.indexOf is not a function. (In 'node.className.indexOf("ion-page")', 'node.className.indexOf' is undefined), http://192.168.2.105:8100/plugins/cordova-plugin-googlemaps/www/Common.js, Line: 237`

wf9a5m75:
Please share your project files on github or bitbacket.

How and why?
#1479

tracedaily:
Same here. It breaks anytime the ionic loading element is used after the map loads. The spinner is svg element, and svg does not have indexOf

wf9a5m75:
Please share your project files on github or bitbacket.

How and why?
#1479

tracedaily:
https://bitbucket.org/tstaley/nativemapsexample

tracedaily:
Click on the 'about' tab, and notice that the button works. Then click on the home tab, and click on the map to add a marker (which displays the loader). You can see the error in the console. Then if you click back on the about tab, the button doesn't work, and any additional taps on any other tab will continue to add more markers to the map, even though it isn't visible

marijnroukens:
Temporary fix untill I find the time to upload the project files:

Replace function `getAllChildren` inside `plugins/cordova-plugin-googlemaps/www/Common.js` with:

```
function getAllChildren(root) {
    if (!root) {
      return [];
    }
    var ignoreTags = ["pre", "textarea", "p", "form", "input", "table", "caption"];

    var list;

    if (window.document.querySelectorAll) {
        // Android: v4.3 and over
        // iOS safari: v9.2 and over
        var childNodes = root.querySelectorAll("*");
        var allClickableElements = Array.prototype.slice.call(childNodes);
        list = allClickableElements.filter(function(node) {
            var tagName = node.tagName.toLowerCase();
            return node !== root && _shouldWatchByNative(node) && ignoreTags.indexOf(tagName) == -1;
        });
    } else {
        var node;
        var clickableElements = root.getElementsByTagName("*");
        for (var i = 0; i < clickableElements.length; i++) {
            node = clickableElements[i];
            if (_shouldWatchByNative(node) && ignoreTags.indexOf(tagName) == -1) {
                list.push(node);
            }
        }
    }

    return list;
}
```

I have not seen any unexpected behaviour so far.

[...]

PR #1504, opened 2017-06-20
Multiple maps

blckshrk:
Hi,

This should solve #1493 by adding `svg` tag to the ignore list. I also refactor the function for the sake of simplicity. This shouldn't change the current behavior.

alex

### DIFF
--- a/www/Common.js
+++ b/www/Common.js
@@ -218,41 +218,30 @@ function getAllChildren(root) {
       return [];
     }
 
-    var ignoreTags = ["pre", "textarea",
-      "p", "form", "input", "table", "caption", "canvas", "svg",
-      "ion-content", "ion-app", "ion-nav"];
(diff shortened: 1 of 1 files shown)


www/Common.js
```diff
@@ -218,41 +218,30 @@ function getAllChildren(root) {
...
-              node.className.indexOf("ion-page") === -1 &&
-              node.className.indexOf("fixed-content") === -1) {
-                list.push(node);
-            }
-        }
+        allClickableElements = root.getElementsByTagName("*");
     }
 
-    return list;
+    return allClickableElements.filter(function (node) {
```
wf9a5m75: The reason of the `for` instead of `filter` is that the webview of Android 4.2 does not support `Array.filter`

blckshrk: Ok, no problem, let's do a for loop. Shouldn't change anything.

blckshrk: Ok, no problem. We can also use a for loop for both. At this level this could make the code more clearer and concise. I can push a for version if you want !

PR #1538, opened 2017-07-09
Fix click events when more than one plugin is active

matthewlloyd:


wf9a5m75:
Could you explain what your pull request is for?

matthewlloyd:
Sure - there is a bug in this block of code. The for loop 

> for (i = 0; i < [pluginNames count]; i++) {

loops over the plugins, and checks each object to see if it overlaps the click point. The ones that do get added to boundsHitList. But then in the for loop

> for (i = 0; i < [boundsHitList count]; i++) {

the value of plugin currently used is the value from the last iteration of the for loop. So, if there is more than one plugin in the list, the objects in the boundsHitList are not found unless the object is from the last plugin, e.g. this line will get nil:

> GMSCircle *circle = (GMSCircle *)[plugin.objects objectForKey:key];

This patch fixes the issue by keeping track of which plugin each matching object came from.

wf9a5m75:
I reviewed the code, and you are correct. Thank you for finding out the critical bug.

PR #1586, opened 2017-07-24
Update PluginTileProvider.m

frank10gm:
`if ([[options objectForKey:@"debug"] boolValue]) {`

I've added "casting" to boolean value, otherwise it's always YES (ionic framework --> for Ionic users, I also added the "debug" option in native plugin)


wf9a5m75:
Oh, thanks.

PR #1613, opened 2017-08-04
refactor(BaseClass): reduces amount of created objects/functions

stalniy:
Utilizes prototype & Symbol instead of closure. This speeds up creation of overlays on map (markers, polygons, etc) and reduces amount of used memory.

Changes:
* implemented EventEmitter completely on JS without DOM events
* Moved all methods into prototype to reduce memory usage
* hid variables and subscriptions under dynamic field name or `Symbol` if available
* deperecated `bindTo`, replaced it with `sync` mehod
* adds `destroy` which calls `off` and `empty`

### DIFF
--- a/www/BaseClass.js
+++ b/www/BaseClass.js
@@ -1,181 +1,130 @@
+var VARS_FIELD = typeof Symbol === 'undefined' ? '__vars' + Date.now() : Symbol.for('vars');
+var SUBSCRIPTIONS_FIELD = typeof Symbol === 'undefined' ? '__subs' + Date.now() : Symbol.for('subscriptions');
 
-var _globalListeners = {};
+function BaseClass() {
+  this[VARS_FIELD] = {};
+  this[SUBSCRIPTIONS_FIELD] = {};
+  this.errorHandler = this.errorHandler.bind(this);
 
-function globalEventListener(e) {
-  var eventName = e.type;
-  if (!_globalListeners[eventName]) {
-    return;
-  }
-
-  var hashCode = e.myself.hashCode;
-  if (!hashCode || !_globalListeners[eventName][hashCode]) {
-    return;
-  }
-  var callbacks = _globalListeners[eventName][hashCode];
-  for (var i = 0; i < callbacks.length; i++) {
-    callbacks[i].listener(e);
-  }
+  Object.defineProperty(this, 'hashCode', { value: Math.floor(Date.now() * Math.random()) })
 }
 
-var BaseClass = function() {
-  var self = this;
-  var _vars = {};
+BaseClass.prototype = {
+  empty: function() {
+    var vars = this[VARS_FIELD];
 
-  var hashCode = Math.floor(Date.now() * Math.random());
+    Object.keys(vars).forEach(function(name) {
+      vars[name] = null;
+    });
+  },
 
-  Object.defineProperty(self, "hashCode", {
-      value: hashCode,
-      writable: false
-  });
-  self.empty = function() {
-    for (var key in Object.keys(_vars)) {
-      _vars[key] = null;
-      delete _vars[key];
-    }
-  };
+  get: function(key) {
+    return this[VARS_FIELD].hasOwnProperty(key) ? this[VARS_FIELD][key] : null;
+  },
 
-  self.get = function(key) {
-    return key in _vars ? _vars[key] : null;
-  };
-  self.set = function(key, value, noNotify) {
-    var prev = _vars[key];
-    _vars[key] = value;
+  set: function(key, value, noNotify) {
+    var prev = this.get(key);
+
+    this[VARS_FIELD][key] = value;
+
     if (!noNotify && prev !== value) {
-      self.trigger(key + "_changed", prev, value);
+      this.trigger(key + '_changed', prev, value);
     }
-  };
-  self.bindTo = function(key, target, targetKey, noNotify) {
-    targetKey = targetKey === undefined || targetKey === null ? key : targetKey;
-    self.on(key + "_changed", function(prevValue, newValue) {
-      target.set(targetKey, newValue, noNotify === true);
+
+    return this;
+  },
+
+  bindTo: function(key, target, targetKey, noNotify) {
+    var targetKey = targetKey || key;
+
+    this.on(key + '_changed', function(oldValue, value) {
+      target.set(targetKey, value, noNotify);
     });
-  };
+  },
 
-  self.trigger = function(eventName) {
-    if (!eventName || !_globalListeners[eventName] || !_globalListeners[eventName][hashCode]) {
-      return;
+  trigger: function(eventName) {
+    if (!eventName || !this[SUBSCRIPTIONS_FIELD][eventName]) {
+      return this;
     }
 
-    var args = [];
-    for (var i = 1; i < arguments.length; i++) {
-      args.push(arguments[i]);
-    }
-    var event = document.createEvent('Event');
-    event.initEvent(eventName, false, false);
-    event.mydata = args;
-    event.myself = self;
+    var listeners = this[SUBSCRIPTIONS_FIELD][eventName];
+    var i = listeners.length;
+    var args = Array.prototype.slice.call(arguments, 1);
 
-    var callbacks = _globalListeners[eventName][hashCode];
-    for (var i = 0; i < callbacks.length; i++) {
-      callbacks[i].listener(event);
-    }
-  };
-  self.on = function(eventName, callback) {
-    if (!eventName || !callback || typeof callback !== "function") {
-      return;
+    while (i--) {
+      listeners[i].apply(this, args);
     }
 
-    _globalListeners[eventName] = _globalListeners[eventName] || {};
+    return this;
+  },
 
-    var listener = function(e) {
-      if (!e.myself || e.myself !== self) {
-        return;
-      }
-      var evt = Object.create(e);
-      var mydata = evt.mydata;
-      delete evt.mydata;
-      delete evt.myself;
-      self.event = evt;
-      callback.apply(self, mydata);
+  on: function(eventName, listener) {
+    this[SUBSCRIPTIONS_FIELD][eventName] = this[SUBSCRIPTIONS_FIELD][eventName] || [];
+    var topic = this[SUBSCRIPTIONS_FIELD][eventName];
+    var index = topic.push(listener);
+    var self = this;
+
+    return function() {
+      self.off(eventName, listener);
     };
-    if (!(hashCode in _globalListeners[eventName])) {
-      document.addEventListener(eventName, globalEventListener, false);
-      _globalListeners[eventName][hashCode] = [];
+  },
+
+  off: function(eventName, listener) {
+    if (!eventName && !listener) {
+      this[SUBSCRIPTIONS_FIELD] = {};
+    } else if (eventName && !listener) {
+      this[SUBSCRIPTIONS_FIELD][eventName] = null;
+    } else if (this[SUBSCRIPTIONS_FIELD][eventName]) {
+      var index = this[SUBSCRIPTIONS_FIELD][eventName].indexOf(listener);
+
+      if (index !== -1) {
+        this[SUBSCRIPTIONS_FIELD][eventName].splice(index, 1);
+      }
     }
-    _globalListeners[eventName][hashCode].push({
-      'callback': callback,
-      'listener': listener
+
+    return this;
+  },
+
+  one: function(eventName, listener) {
+    var unlisten = this.on(eventName, function() {
+      unlisten();
+      listener.apply(this, arguments);
     });
-  };
-  self.addEventListener = self.on;
 
-  self.off = function(eventName, callback) {
-    var i, j, callbacks;
-    if (typeof eventName === "string") {
-      if (eventName in _globalListeners) {
+    return this;
+  },
 
-        if (typeof callback === "function") {
-          callbacks = _globalListeners[eventName][hashCode] ||[];
-          for (i = 0; i < callbacks.length; i++) {
-            if (callbacks[i].callback === callback) {
-              callbacks.splice(i, 1);
-              break;
-            }
-          }
-          if (callbacks.length === 0) {
-            delete _globalListeners[eventName][hashCode];
-          }
-        } else {
-          delete _globalListeners[eventName][hashCode];
-        }
-        if (Object.keys(_globalListeners[eventName]) === 0) {
-          document.removeEventListener(eventName, globalEventListener);
-        }
-      }
-    } else {
-      //Remove all event listeners
-      var eventNames = Object.keys(_globalListeners);
-      for (j = 0; j < eventNames.length; j++) {
-        eventName = eventNames[j];
-        delete _globalListeners[eventName][hashCode];
-        if (Object.keys(_globalListeners[eventName]) === 0) {
-          document.removeEventListener(eventName, globalEventListener);
-        }
-      }
+  destroy: function() {
+    this.off();
+    this.empty();
+  },
+
+  errorHandler: function(error) {
+    if (error) {
+      console.log(error);
+      this.trigger('error', error instanceof Error ? error : createError(error));
     }
-  };
 
-  self.removeEventListener = self.off;
-
-
-  self.one = function(eventName, callback) {
-
-    var listener = function(e) {
-      if (!e.myself || e.myself !== self) {
-        return;
-      }
-      var evt = Object.create(e);
-      var mydata = evt.mydata;
-      delete evt.mydata;
-      delete evt.myself;
-      self.event = evt;
-      callback.apply(self, mydata);
-      self.off(eventName, callback);
-    };
-
-    _globalListeners[eventName] = _globalListeners[eventName] || {};
-
-    if (!(hashCode in _globalListeners[eventName])) {
-      document.addEventListener(eventName, globalEventListener, false);
-      _globalListeners[eventName][hashCode] = [];
-    }
-    _globalListeners[eventName][hashCode].push({
-      'callback': callback,
-      'listener': listener
-    });
-  };
-  self.addEventListenerOnce = self.one;
-
-  self.errorHandler = function(msg) {
-    if (msg) {
-      console.log(msg);
-      self.trigger('error', msg);
-    }
     return false;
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Please provide more information. The prof of reduce memory and improve speed up. 
Also test code.

And why did you remove the hash code property?
You should not remove the property currently provided.

stalniy:
I haven't removed `hashCode`:

```js
Object.defineProperty(this, 'hashCode', { value: Math.floor(Date.now() * Math.random()) })
```

wf9a5m75:
And also you should not change the method name, bindTo().
I designed this plugin similar with the Google Maps JavaScript v3.

And also did you test this code does not affect any current code, even multiple maps?
If you did, please provide how did you test?
This only JS code, and BaseClass effects or affects all classes, so you are trying to make a big change.

I believe you created safe test code a lot.

I can not accept your pull request soon. Please provide more information.
At least, I decline to change the name of method from bindTo to sync.

wf9a5m75:
The hashcode property is my mistake.

Did you create unit test, of course?
I asked please provide the unit test code you created.

BaseClass is just JS code. You can test only this class on browser, of Node.js.

stalniy:
I did brief testing. This is the most simple event emitter. However I will add unit tests for `BaseClass`

wf9a5m75:
Test is really important especially for pull request.

In general (I'm not pointting only you),
from the sender a pull request side, if it is accepted, that's all.
But if it has a bug and reproduce is difficult, who has responsibility?
The person who accept the pull request, it is me.
In order to confirm and fix the bug, I need lots of spend time.

Especially you are trying to make a big change.
That's why I ask the proof (such as screen captures of memory usage of Chrome Developer tool and safari inspector), and unit test code.

Unit test has to include two side:
- the proof of improvement
- and the proof of safe that your change does not affect any current code

Thank you for your preparation.

stalniy:
I know that it's important but I don't see any tests in this repository ;)

wf9a5m75:
I do in my local, but I can not share them. If you send me a more public test, I appreciate for you.

wf9a5m75:
By the way, you should not use the name as `Cordova GoogleMaps plugin for iOS and Android (version 3.0)`.

Because you are not official. Since this plugin adapts the Apache License 2.0, I don't mind if you change the code, and redistribute it, but plugin name is not good.
At least you should change the name.
https://github.com/stalniy/cordova-plugin-googlemaps

wf9a5m75:
And one more by the way.
Don't misunderstand I don't refuse your pull request.
I just worry about your pull request.
Improve pull request is appreciate.
That's why I open the full source code.

stalniy:
I know but you should understand me as well: I don't want to spend a lot of time doing screenshots and calculating metrics. Better to spend it optimizing code, adding features or tests, real unit tests :)

wf9a5m75:
As I wrote, your code effect or affect the code overall.
Please recognize you are trying to make a big change to this plugin.
I understand you don't want to spend time, but I really worry about your pull request. Because the impact is so big.
That's why I'm asking.

If this pull request improves the performance significantly, I will pay some amount for your work.
This is the correct way of the project donation. (Don't expect too much. The plugin project is poor.)

So what I want to know:
- your code is safe or not,
- how improve the performance

If you are tired today, you can do later.
I have a high expectation for your work ;)

wf9a5m75:
I got an error with this simple code... 
```js
var div = document.getElementById("map_canvas");
var map = plugin.google.maps.Map.getMap(div);
map.one(plugin.google.maps.event.MAP_READY, function() {
  var htmlInfoWindow = new plugin.google.maps.HtmlInfoWindow();

  var html = [
    'This is &lt;b&gt;Html&lt;/b&gt; InfoWindow',
    '&lt;br&gt;',
    '&lt;button onclick="javascript:alert(\'clicked!\');"&gt;click here&lt;/button&gt;',
  ].join("");
  htmlInfoWindow.setContent(html);

  map.addMarker({
    position: {lat: 0, lng: 0},
    draggable: true
  }, function(marker) {

    marker.on(plugin.google.maps.event.MARKER_CLICK, function() {
      htmlInfoWindow.open(marker);
    });
    marker.trigger(plugin.google.maps.event.MARKER_CLICK);

  });
});
```


screen shot 2017-08-04 at 3 11 39 pm

----

**at HTMLInfoWindow.trigger (BaseClass.js:64)**
screen shot 2017-08-04 at 3 14 44 pm

**at HTMLInfoWindow.<anonymous> (HtmlInfoWindow.js:170)**
screen shot 2017-08-04 at 3 15 59 pm


**at HTMLInfoWindow.trigger (BaseClass.js:64)**
screen shot 2017-08-04 at 3 14 44 pm

**at HTMLInfoWindow.calculate (HtmlInfoWindow.js:154)**
screen shot 2017-08-04 at 3 15 29 pm


**at HTMLInfoWindow.trigger (BaseClass.js:64)**
screen shot 2017-08-04 at 3 14 44 pm

**at Map.<anonymous> (HtmlInfoWindow.js:247)**
screen shot 2017-08-04 at 3 14 29 pm

**at Map.js:648**
screen shot 2017-08-04 at 3 18 29 pm

**at googlemaps-cdv-plugin.js:424**
screen shot 2017-08-04 at 3 19 17 pm

wf9a5m75:
And do not remove the `addEventListener`, `addEventListenerOnce`, and `removeEventListener` methods.

As I wrote, I designed this plugin is similar with the Google Maps JavaScript API v3.
The reason of existing `on()` and `addEventListener()` methods are I want to bring down the introduction step wall for beginners.

`on()` method is really popular, but also some pure JavaScript users prefer `addEventListener()`.
So I decided I provide both, even it is alias.
Did you consider why I provide both methods?

I think I wrote to you before in another thread, please consider `why this code is existed`.
Keeping backward compatibility is really important for this plugin. Do not change the method name, do not remove methods please.
(Except very early implement, such as the method was implemented just one day ago.)

wf9a5m75:
Since you removed `addEventListener()`, this code does not work.

```js
map.addEventListener(plugin.google.maps.event.MAP_READY, function() {

});
```

----

As you can see, if you think `oh, I prefer on() instead of addEventListener(), let's remove it`.
Then code is break.

Making a change to BaseClass means you are trying to make very big impact.

That's why I'm asking `please make test code`.

wf9a5m75:
I checked half of examples of [the official demo app](https://github.com/mapsplugin/v2.0-demo)
As I tested, **mostly fine except `HTMLInfoWindow`**.

I haven't checked performance.
You described your code improve the performance , reduce memory usage.
Please make a proof.
For example, adding 1,000 markers, how many memory the current code, and your code?

Without the proof and fixing the error of HTMLInfoWindow, I can not accept your pull request.
In other word, I will accept you resolve the issue and make the proof.
Then I pay some amount for your work.

stalniy:
fair enough. I will check that use case, add it to unit tests and will update PR. Thanks for the help with testing!

I will add back `addEventListener` stuff and `bindTo`

stalniy:
@wf9a5m75 by the way, thanks for suggestion to pay me but I don't need this. Just include my nickname in release notes and this will be enough. Thanks

stalniy:
I found the issue, it can be fixed on `BaseClass` level but I believe this is `HTMLInfoWindow` issue  and should be fixed there.

Check https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/multiple_maps/www/HtmlInfoWindow.js#L245

You will find:
```js
self.on(event.INFO_OPEN);
```
So, the question why is `on` called without listener? Looks like a bug, am I right?

stalniy:
I used `newHtmlInfoWindow.html` demo with 1000 markers (you can generate markers using http://www.geomidpoint.com/random/ + regexp find/replace to convert it to JSON).

Timeline for old `BaseClass`:
timeline

Timeline for refactored `BaseClass`:
timeline-new

As you can see **Scripting** part for refactored `BaseClass` is about on 400-500 ms lower.

Hope this is enought to proof that this code is  faster.

wf9a5m75:
Thank you for testing.
I just waked up, so I haven't checked this yet.

wf9a5m75:
I executed 16077 times of `new Marker()` (not `map.addMarker`).

This is current code.

screen shot 2017-08-05 at 4 26 35 pm

And your code.

screen shot 2017-08-05 at 4 20 25 pm

That's amazing! Almost 60% faster.
(Memory is actually increased though, not too much. So it's ok)

---

I will test all methods on Android and iOS.
Please wait for a while.

wf9a5m75:
And do you have a paypal account? If no, please tell me the best way to receive some amount for you?

Please send me the information to my e-mail address.

www/BaseClass.js
```diff
@@ -1,181 +1,130 @@
...
+    var vars = this[VARS_FIELD];
 
-  var hashCode = Math.floor(Date.now() * Math.random());
+    Object.keys(vars).forEach(function(name) {
+      vars[name] = null;
+    });
+  },
+
+  get: function(key) {
+    return this[VARS_FIELD].hasOwnProperty(key) ? this[VARS_FIELD][key] : null;
```
wf9a5m75: This should be `return this[VARS_FIELD].hasOwnProperty(key) ? this[VARS_FIELD][key] : undefined`.
The null === undefined always false.

wf9a5m75:
I sent you some amount to your e-mail from my paypal account. Please follow the instruction.
Thank you for improving the cordova-googlemaps-plugin significantly.

www/BaseClass.js
```diff
@@ -1,181 +1,130 @@
...
-      delete evt.myself;
-      self.event = evt;
-      callback.apply(self, mydata);
+  on: function(eventName, listener) {
+    this[SUBSCRIPTIONS_FIELD][eventName] = this[SUBSCRIPTIONS_FIELD][eventName] || [];
+    var topic = this[SUBSCRIPTIONS_FIELD][eventName];
+    var index = topic.push(listener);
+    var self = this;
+
+    return function() {
```
wf9a5m75: Could you tell me why do you return this function? 
```
function() {
  self.off(eventName, listener)
}
```

I think this is not common way...
```js
map.on(plugin.google.maps.event.MAP_READY, onMapReady)();
```

stalniy:
it's much easier to work with subscriptions than with callbacks.

Lets imagine that you need to subscribe some function change event and afterwards when component or object is destroyed - unsubscribe it. Then you will do something like this:

```js
class Component {
  constructor(input) {
     this.input = input
     this.changeListener = () => { console.log('changed!') }
     this.input.on('change', this.changeListener)
  }

  onDestroy() {
    this.input.off('change', this.changeListener)
  }
}
```
You need to store listener function in class instance. It's not very convenient especially when your function is big or just a wrapper for another function (e.g., `(event) => { this.isClicked(event.pageX, event.pageY) }`). 

Instead it's more convenient (and more logically) to store reference to a `subscription` instance and eventually dispose subscription:

```js
class Component {
  constructor(input) {
     this.input = input
     this.unsubscribe = this.input.on('change', () => { console.log('changed!') })
  }

  onDestroy() {
    this.unsubscribe()
  }
}
```

Code is cleaner.

Who uses such appoach:
* [rxjs](http://reactivex.io/rxjs/manual/overview.html#disposing-observable-executions) (used by Angular 2+ and Ionic 2+)
* [knockoutjs](http://knockoutjs.com/documentation/observables.html#explicitly-subscribing-to-observables)

wf9a5m75:
This is not the way I designed.
Since your code blocks on my way, I can not choose to support `Promise`,

wf9a5m75:
In order to unsubscribe easily, I provide `obj.off()`, `obj.off("eventName")` and `obj.off("eventName", callback)`.
This is enough.

wf9a5m75:
Sorry, but according to Google Trends, rxjs and knockoutjs are not so popular choice.

**Past 12 months**
screen shot 2017-08-08 at 2 27 30 pm

https://trends.google.com/trends/explore?q=ionic,rxjs,knockoutjs,cordova

**Past 5 years**
screen shot 2017-08-08 at 2 29 02 pm

https://trends.google.com/trends/explore?date=today%205-y&q=ionic,rxjs,knockoutjs,cordova


So if I am forced to choose make a choice, I definitely choose `Promise` because this is ionic way.
And `Promise` is almost work on every devices.
(One year ago was different. That's why I did not choose using Promise)

wf9a5m75:
Actually I have to make a choice to support the ionic-native/google-maps wrapper plugin.

stalniy:
How do you want to use `Promise`?

I agree about knockout, but rxjs became popular when Angular2 was release and Ionic2 is built on top of Angular2.

wf9a5m75:
The problem of the wrapper plugin is which uses `then` and `subscribe`.

For example:

```js
this.map.one(GoogleMapsEvent.READY, () => {

   for (let i = 0; i < locations.length; i++) {
     this.map.addMarker(locations[i], this.onMarkerAdded);
   }

});

onMarkerAdded(marker: Marker) {
  marker.on(GoogleMapsEvent.MARKER_CLICK, this.onMarkerClick);
}

onMarkerClick(...params: any[]) {
  let latLng = params[0];
  let marker = params[1];
  marker.showInfoWindow();
}
```

In order to use write like the above code, `on` and `one` have to return the object of the class of the wrapper plugin.

But `on` and `one` of the wrapper plugin passes the raw values to the event lister.

-------

Currently, the JS plugin passes the marker instance for all marker events.
It is not written in the document though.
Because ionic does not allow to use `this` to obtain the caller object of event listener.

**Ionic does not allow this way**
```
marker1.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);
marker2.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);

onMarkerClick() => {
  // there is no way to detect which marker called this event listener
}
```

In order to solve this problem, the plugin passes the instance as the second argument.
(But current ionic wrapper plugin does not receive it though. I'm patching.)

**goal**

```
marker1.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);
marker2.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);

onMarkerClick(...params: any[]) => {
  let latLng = params[0];
  let marker = params[1];
}
```

-------

However, this plugin does not support `Promise (browser native)` and `Observable (rxjs)`,
the wrapper plugin has to convert the instance type from pure JS to the ionic class,
something like this (this code might not work, just example):

```
  on(eventName: string): Observable<any> {

    return  new Observable((observer) => {
         this._objectInstance.on(eventName, (...results: any[]) {
           for (let i = 0; i < results.length; i++) {
             if (results[i] instanceof GoogleMaps.getPlugin().BaseClass) {
               results[i] = convertFromJS_to_Ionic(results[i]));
             }
           }
           observer.next.apply(observer, results);
         });
       }
     );
     return obs;
  }
```

Otherwise, this code would become like this:

```
marker1.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);
marker2.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);

onMarkerClick(...params: any[]) => {
  let latLng = params[0];
  let marker = params[1];

  // This does not work, because marker is instance of JS, not ionic wrapper's one.
  // marker.one(GoogleMapsEvent.INFO_CLICK).then(() => {
  //   alert("info clicked");
  // });

  // This works, but this leads many people confuse a lot.
  marker.one(GoogleMapsEvent.INFO_CLICK, () => {
    alert("info clicked");
  });
}
```

-----------

So I'm planing to use the `Promise` and `Observable` if these both classes are available.
If the JS side supports them, the ionic wrapper plugin does not have to convert the instance.

```
marker1.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);
marker2.on(GoogleMapsEvent.MARKER_CLICK, onMarkerClick);

onMarkerClick(...params: any[]) => {
  let latLng = params[0];
  let marker = params[1];

  // This code works if the JS side support Promise and Observable, even if the marker instance is JS instance.
  marker.one(GoogleMapsEvent.INFO_CLICK).then(() => {
     alert("info clicked");
  });
}
```

The ionic wrapper plugin keeps the same method names as the JS plugin.
So you don't need to keep in mind the instance is JS or ionic one.

----------

I need to detect the Observable is available or not, but if the JS plugin supports Promise and Observable both, the benefit is big for ionic users.
And reduce problems for me.

Hence I need to return something:

```
  on: function(eventName, listener) {
    this[SUBSCRIPTIONS_FIELD][eventName] = this[SUBSCRIPTIONS_FIELD][eventName] || [];
    var topic = this[SUBSCRIPTIONS_FIELD][eventName];
    var index = topic.push(listener);
    var self = this;

    if (checkAvailabilityObservable()) {
       return new Observable(...);
    } else {
       return this;
    }
  },
```

-------------

And you may notice that the JS plugin currently returns their instance of the class.

https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/multiple_maps/www/Marker.js#L231

```js
Marker.prototype.setFlat = function(flat) {
    flat = common.parseBoolean(flat);
    this.set('flat', flat);
    return this;
};
Marker.prototype.setIcon = function(url) {
    if (url && common.isHTMLColorString(url)) {
        url = common.HTMLColor2RGBA(url);
    }
    this.set('icon', url);
    return this;
};
```

I designed this allows you to use the chain method like jquery.

```
marker.setTitle("HelloWorld").showInfoWindow();
```

I forgot to return `this` in the `on` method, but I prefer this way.

```
marker.on(plugin.google.maps.event.MARKER_CLICK, onMarkerClick)
            .on(plugin.google.maps.event.INFO_CLICK, onInfoWndClick);
```

-----------

I understand your point, but as I write previously, this plugin supports three ways `obj.off()`, `obj.off("eventName")`, and `obj.off("eventName", callback)`.

You just keep the callback if you want to remove particular event listener.

```
class Component {
  constructor(input) {
     this.input = input
     this.change = () => { console.log('changed!') };
  }

  onDestroy() {
    this.input.on('change', this.change);
  }
}
```

or simply just `off()` or `destroy (you added)`

```
class Component {
  constructor(input) {
     this.input = input
     this.change = () => { console.log('changed!') };
  }

  onDestroy() {
    this.input.destroy();
  }
}
```

wf9a5m75:
And this code create different closures every time.

```
return function() {
      self.off(eventName, listener);
    };
```

It means if you create 1,000 markers, this code creates another 1,000 closures.

Did you say this is bad performance, didn't you?

stalniy:
ok, it makes sense

PR #1615, opened 2017-08-05
fix(map): fixes issue with invalid string length when options.styles is specified

stalniy:
when you specify `options.styles` for the map the they are converted several times into `JSON` and eventually JS code raises exception `Invalid string length`. The bug is produced because code changes passed in `options` argument instead of making a copy of `options` and change a copy.
 (https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/multiple_maps/www/Map.js#L123)

This is hot fix which checks that `options.styles` is array and only then converts it into JSON


wf9a5m75:
Thanks

PR #1618, opened 2017-08-06
fix(browserify): removes `return` statement which was used outside of function scope

### LINKED ISSUE #1616: Doesn't work with `--browserify` option
**I'm submitting a ...**  (check one with "x")
[ ] question
[x] any problem or bug report
[ ] feature request

**plugin version:**  (check one with "x")
[ ] **1.4.x**
[x] **2.0.0-beta2** 

**Current behavior:**
```sh
$ cordova build ios --emulator --browserify
Error: 'return' outside of function (5:2) (while aliasify was processing /home/sergii/projects/misto/plugins/cdv-googlemaps/www/googlemaps-cdv-plugin.js) while parsing file: /home/sergii/projects/misto/plugins/cdv-googlemaps/www/googlemaps-cdv-plugin.js
```
**Expected behavior**:

it should be possible to create browserified build

stalniy:
This fixes `cordova build --browserify` build and allows to improve app boot time significantly!

Fixes #1616

### DIFF
--- a/www/googlemaps-cdv-plugin.js
+++ b/www/googlemaps-cdv-plugin.js
@@ -1,9 +1,5 @@
 /* global cordova, plugin, CSSPrimitiveValue */
 var MAP_CNT = 0;
-
-if (!cordova) {
-  return;
-}
 var argscheck = require('cordova/argscheck'),
     utils = require('cordova/utils'),
     cordova_exec = require('cordova/exec'),


wf9a5m75:
This return function is important.

When cordova executes the app, cordova wraps with a function.
This return statement prevents loading the script that is no cordova script.

stalniy:
Do you have a use case for this script to be run outside of cordova?

wf9a5m75:
Yes.

https://github.com/mapsplugin/v2.0-demo/blob/master/www/Map/index.html

wf9a5m75:
I see many apps, not only pure cordova app, but also include many frameworks.

For example, DrupalGap is a framework that uses cordova.
But the users of this framework mostly create thier apps like webpage.
Only the map page uses cordova with this plugin.

stalniy:
if you open cordova app in browser all cordova plugins won't be included. Also it doesn't make much sense because this plugin won't work without cordova env at all, isn't it?

stalniy:
I did quick grep over this plugin and it shows that there are 15 files which requires `cordova` or cordova related modules. So, all that files may be run only in cordova env.

wf9a5m75:
No, you are not correct.

The pages does not include <script src="cordova.js"> still loads my plugin.
Because I specify to do that in the plugin.xml
(I just waked up, so I forgot which one is exactly that)

Because of this plugin fires the plugin_touch event from the native side, the cordova is undefined error at the these pages.
And also this plugin changes the page viewport, hook the backbutton, etc.
I don't want to run the script except maps page.

stalniy:
ok, so we need to find a different solution... because this blocks to improve app load time with browserify... Read my comment for details - https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/1616#issuecomment-320498804

PR #1630, opened 2017-08-10
Fixes iOS 11 white statusbar issue in Cordova

longzheng:
iOS 11 changes the behaviour of how UIWebView and WKWebView calculates the layout. 

<iOS 11 it would return the layout including the status bar so you could render underneath it. 
In IOS 11 by default the layout no longer includes the status bar (which is then rendered as white).

As the Cordova team has suggested, you can add a viewport tag ```viewport-fit=cover``` to change to the intended behaviour of including the status bar. https://issues.apache.org/jira/browse/CB-12886

However this plugin overrides the viewport tag on load with its own custom value. This corrects the value to also include the fix.

### DIFF
--- a/www/googlemaps-cdv-plugin.js
+++ b/www/googlemaps-cdv-plugin.js
@@ -59,7 +59,7 @@ if (!cordova) {
         viewportTag = document.createElement("meta");
         viewportTag.setAttribute('name', 'viewport');
     }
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Thank you for letting me know.
Actually I haven't tested this plugin on iOS11 yet.

wf9a5m75:
https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/1637

PR #1753, opened 2017-09-19
CSS Classes. Clear content before append.

superpowered:
Adds CSS classes to the divs generated for the info window. currently the only way to target them is like 
```
#map
{
     & > div:first-of-type
     {
        & > div:first-of-type
         {
         }
     }
}

```
which is not great.

The other bit is just clearing the content of the contentbox before appending new content, there was a strange bug that surfaced over the weekend where boxes would stack on top of one another instead of replacing themselves if you used HTML elements

### DIFF
--- a/www/HtmlInfoWindow.js
+++ b/www/HtmlInfoWindow.js
@@ -15,11 +15,13 @@ var HTMLInfoWindow = function() {
     frame.style.overflow="visible";
     frame.style.position="absolute";
     frame.style.display = "inline-block";
+    frame.classList.add('gm-html-info-frame');
     self.set("frame", frame);
 
     var contentBox = document.createElement("div");
     contentBox.style.display = "inline-block";
     contentBox.style.padding = "5px";
+    contentBox.classList.add('gm-html-info-content-box');
 
     var contentFrame = document.createElement("div");
     contentFrame.style.display = "block";
@@ -29,12 +31,14 @@ var HTMLInfoWindow = function() {
     contentFrame.style.left = "0px";
     contentFrame.style.right = "0px";
     contentFrame.style.zIndex = "1";  // In order to set higher depth than the map div certainly
+    contentFrame.classList.add('gm-html-info-content-frame');
     frame.appendChild(contentFrame);
     contentFrame.appendChild(contentBox);
 
     var tailFrame = document.createElement("div");
     tailFrame.style.position = "relative";
     tailFrame.style.marginTop = "-1px";
+    tailFrame.classList.add('gm-html-info-tail-frame');
     frame.appendChild(tailFrame);
 
     var tailLeft = document.createElement("div");
@@ -82,6 +86,7 @@ var HTMLInfoWindow = function() {
           if (!content) {
             contentBox.innerText = "";
           } else if (content.nodeType === 1) {
+            contentBox.innerHTML = "";
             contentBox.appendChild(content);
           } else {
             contentBox.innerText = content;


wf9a5m75:
```css
#myContents h3 {
  color: red;
}
```
```js
var html = "<div id='myContents'><h3>HelloWorld</h3></div>";
var htmlInfoWindow = new plugin.google.maps.HtmlInfoWindow();
htmlInfoWindow.setContent(html);
htmlInfoWindow.open(marker);
```

superpowered:
I know you can add an ID into setContent, however that doesn't allow you to style the frame itself.

This adds a class higher up. In your example, the DOM tree looks like this:
```
<div id='map'>
    <div>
        <div>
            <div>
                <div  id='myContents'>
                    <h3>HelloWorld</h3>
                </div>
            <div>
        </div>
        <div>
            <div></div>
            <div></div>
        <div>
    </div>
</div>
```
adding the classes in would change that same code to:
```
<div id='map'>
    <div class='gm-html-info-frame'>
        <div class='gm-html-info-content-box'>
            <div class=''gm-html-info-content-frame''>
                <div id='myContents'>
                    <h3>HelloWorld</h3>
                </div>
            <div>
        </div>
        <div class=''gm-html-info-tail-frame''>
            <div></div>
            <div></div>
        <div>
    </div>
</div>
```

This allows a lot more flexibility and clarity in styling without having to target a bunch of nameless divs.
```
#map > div:first-of-type > div > div
{
} 
```
vs,
```
#map .gm-html-info-frame
{
}
```

superpowered:
Also:
This would allow in a future update to possibly move some of the hardcoded styling to either an external sheet, or an inline style block. Currently the only way to override the padding and borders, for example, is to use !important at the end of the CSS declaration.

wf9a5m75:
Changing the padding and borders affect the position calculation. Not acceptable.

superpowered:
That's more of an example for user preference. 
I, for example, stripped all of the default styling (borders, padding) and still got the position to work fine, but there was a lot of work in brute forcing my styles to appear.


```
#map,
    agm-map
    {
        height: 100%;

        & > div:first-of-type
        {
            transition: opacity .2s, transform .5s;
            transform: translateY(-20px);
            & > div:first-of-type
            {
                border: none !important;
                -webkit-box-shadow: 5px 5px 26px 0 rgba(0,0,0,0.75);
                -moz-box-shadow: 5px 5px 26px 0 rgba(0,0,0,0.75);
                box-shadow: 5px 5px 26px 0 rgba(0,0,0,0.75);
                position: relative;
                height: auto !important;
                max-width: 150px;
                margin: 0 auto;
                &:after
                {
                    content: '';
                    display: block;
                    height: 3rem;
                    width: 3rem;
                    position: absolute;
                    top: calc(100% - 2rem);
                    left: 50%;
                    transform: translateX(-50%) rotate(45deg);
                    background-color: inherit;
                    z-index: 1;
                    -webkit-box-shadow: 7px 7px 5px 0 rgba(0, 0, 0, 0.15);
                    -moz-box-shadow: 7px 7px 5px 0 rgba(0, 0, 0, 0.15);
                    box-shadow: 7px 7px 5px 0 rgba(0, 0, 0, 0.15);
                }
                & > div
                {
                    position: relative;
                    padding: 0 !important;
                    z-index: 2;
                    height: 100%;
                    width: 100%;
                    top: 0;
                    left: 0;
                    & > div // Here's where setContent starts
                    {
                        position: relative;
                        top: 0;
                        left: 0;
                        height: 100%;
                        width: 100%;
                        img
                        {
                            width: 100%;
                            display: block;
                            background-color: orange;
                            min-height: 2rem;
                        }
                        div
                        {
                            padding: 1rem 2rem 1rem 1rem;
                            color: white;
                            text-align: center;
                            font-size: .875em;
                            line-height: 1.25;
                            position: relative;
                            &:after
                            {
                                content: '';
                                display: block;
                                position: absolute;
                                right: 1rem;
                                top: 50%;
                                width: .5em;
                                height: .5em;
                                border-top: solid 1px white;
                                border-right: solid 1px white;
                                -webkit-transform: translateY(-50%) rotate(45deg);
                                transform: translateY(-50%) rotate(45deg)
                            }
                            span
                            {
                                font-weight: bold;
                            }
                        }
                    }
                }
            }
            & > div:nth-of-type(2)
            {
                display: none !important;
                & > div:first-of-type
                {

                }
                & > div:nth-of-type(2)
                {

                }
            }
        }
        &.info-open
        {
            & > div:first-of-type
            {
                opacity: 0;
                transform: translateY(-40px);
            }
        }
    }
```
creates:

img_3490

But notice how convoluted the styling has to get in order to target the frame and such and how many !important tags need to be included.

Understandably things like height, width, top& left should be added in JS, but likely the other more decorative styles could be better handled in a style block. Even then, this PR isn't even trying to stretch that far. More just a quality of life improvement for anyone trying to do more than a default white window.

wf9a5m75:
Did you consider something like this case?

```css
.gm-html-info-frame {
  border-left: 1px solid red;
  border-top: 1px solid red;
  border-right: 5px solid #AA0000;
  border-bottom: 5px solid #AA0000;
}

```
screen shot 2017-09-19 at 3 16 54 pm

wf9a5m75:
I understand your point, such as border less infoWindow.
But giving CSS class means giving freedom for users, it means easy to break the infoWindow.
Even the Google Maps JavaScript API v3 does not allow it.

wf9a5m75:
The [line 89](https://github.com/mapsplugin/cordova-plugin-googlemaps/pull/1753/files#diff-64653469ccf8a447e9791f08a44c0e2dR89), thank you for your pointing out.

wf9a5m75:
Don't misunderstand. Until the close this PR, I haven't refused your PR. I'm researching.

superpowered:
Yeah, one issue with it is that it does give people the ability to make something odd and broken like that. But even then, that's on the individual, I would argue. It just feels less natural to style a bunch of `div > div` blocks than a class dedicated to the specific bit.

The Javascript API does add some top level classes:
 `gm-style-iw-outer` and `gm-style-iw`.
Although, I've always thought the could add a few more classes. (they still have 2~3 levels of empty divs) I've always felt the infowindow was kind of an underdeveloped feature in the Javascript API.

wf9a5m75:
`gm-style-iw` and `gm-style-iw-outer` are not official documented class names.
It means Google is able to change the class names without notifications.

wf9a5m75:
And if Google uses `gm-style-xxx`, we should not use the `gm-html-xxx`.
It might lead people to confuse.

superpowered:
True, true. I chose that naming convention pretty hastily. You could use whatever feels most appropriate. I didn't want to match it _perfectly_ to google's however since the infowindows are constructed in a different format.

google uses the `gm-style-iw-outer` to style the "bubble" and then the `gm-style-iw` is positioned in front of it, where here we have a DOM tree that drills straight down with the exception of the "tail".

wf9a5m75:
I accepted your idea, but changing the style is used on the users responsibility.
I might change the class name or HTML structure without notification if I need to change them.

PR #1787, opened 2017-09-30
error loading in external maps

blueprin4:
The map was loading latlng as {"lat":xxx,"lng":xxx) and saying location could not be found.  Removed "lat","lng"

### DIFF
--- a/www/LatLng.js
+++ b/www/LatLng.js
@@ -34,7 +34,7 @@ LatLng.prototype = {
    * @return {String} latitude,lontitude
    */
   toString: function() {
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Not acceptable. And external map feature is deprecated. The code will be in the next release.

blueprin4:
Hi Sir after changing www/LatLng.js my app works now.  The way is was before it was passing "lat" and "lng" along with the coordinates so maps or google maps will not load. Here is screenshot of before: https://ibb.co/i4bAG5

wf9a5m75:
No more external feature.

PR #1794, opened 2017-10-03
fix: overlays click not working in iOS

watacoz3:
overlay object was not found when clicking because overlay and its property ids are generated in different ways.


wf9a5m75:
Thank you. This is merged into the multiple_maps branch.

watacoz3:
thank you!

PR #1804, opened 2017-10-10
fix define var `self` in `remove` method of overlays

watacoz3:
there are times when native code not called when removing map.  
I could not catch up to details, but I guessed this to be the cause.  
by the way, `self` refers to `window`.

### DIFF
--- a/www/Circle.js
+++ b/www/Circle.js
@@ -143,9 +143,10 @@ Circle.prototype.setRadius = function(radius) {
 };
 
 Circle.prototype.remove = function() {
+    var self = this;
     this.trigger(this.id + "_remove");
     exec.call(this, null, this.errorHandler, this.getPluginName(), 'remove', [this.getId()]);
-    Object.defineProperty(this, "_isRemoved", {
+    Object.defineProperty(self, "_isRemoved", {
         value: true,
         writable: false
     });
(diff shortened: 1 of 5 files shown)


wf9a5m75:
Thank you for pointing out, it is mistake. But you misunderstand the code.
The code have to be like this.

```js
Circle.prototype.remove = function() {
    var self = this;
     this.trigger(this.id + "_remove");
     exec.call(this, null, this.errorHandler, this.getPluginName(), 'remove', [this.getId()]);
    Object.defineProperty(self, "_isRemoved",
         value: true,
         writable: false
     });
```

watacoz3:
I think it is unnecessary to define `self`, but why is it necessary?  
`this` reference is always same in this method.

wf9a5m75:
@watacoz3 Sorry, I misunderstood. I first checked on my phone, so I thought the code was in callback.

Anyway, I make a point to use the variable `self` instead of `this` usually (except short code).

The reason of avoiding `this` is that clarify `who is this?`.
The keyword `this` indicates who called this function, but it leads to misunderstand people `who is this`.
That's why I use the variable `self` basically, except short code (such as just a few lines).

The code of `remove` methods are mistake of mine.
I copied the code from the `map.js` (I think I was tired), then I didn't notice that there is no definition `self`.
Thank you for pointing out.

I already fixed the code on the multiple_maps branch.

watacoz3:
thank you for your answer and fix.
certainly handling `this` is easy to mistake in JS.
I think that it is good to do so.

PR #1875, opened 2017-11-12
Update MyPluginLayer.m

z1ad:
fix issue with not being able to interact with any other view that might have been added
e.g. PhoneGap-Plugin-ListPicker, cordova-plugin-datepicker, etc

Note: this issue only affects iOS. Android already works fine

### DIFF
--- a/src/ios/GoogleMaps/MyPluginLayer.m
+++ b/src/ios/GoogleMaps/MyPluginLayer.m
@@ -294,26 +294,25 @@ BOOL hasCordovaStatusBar = NO;  // YES if the app has cordova-plugin-statusbar
     }
 }
 - (UIView *)hitTest:(CGPoint)point withEvent:(UIEvent *)event {
+    // Check other views of other plugins before this plugin
+    // e.g. PhoneGap-Plugin-ListPicker, etc
+    for (UIView *subview in [self.webView.superview subviews]) {
+      // we only want to check against other views
+      if (subview == self.pluginScrollView || subview == self.webView) {
+        continue;
+      }
+
+      UIView *hit = [subview hitTest:point withEvent:event];
+
+      if (hit) {
+        return hit;
+      }
+    }
+  
     if (self.isSuspended || self.pluginScrollView.debugView.mapCtrls == nil || self.pluginScrollView.debugView.mapCtrls.count == 0) {
       // Assumes all touches for the browser
       [self execJS:@"javascript:if(window.cordova){cordova.fireDocumentEvent('plugin_touch', {});}"];
-
-	  // fix issue with not being able to interact with any other view that might have been added
-	  // e.g. PhoneGap-Plugin-ListPicker, etc
-	  for (UIView *subview in [self.webView.superview subviews])
-	  {
-		  // we only want to check against other views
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Thank you for sending a PR. Please create a demo project, then share it to GitHub.

VeteranCoder:
Hello,

I just wanted to add my two cents to this thread.  We're having the same issue here with the cordova-plugin-datepicker, and are eager to see a fix implemented to this plugin.  Do you have any sense of when the fix might be accepted and included in a release?

wf9a5m75:
If someone provide enough test project on github, I will check it.

z1ad:
https://github.com/z1ad/cordova-plugin-googlemaps/blob/master/test/index.html

cordova plugin add cordova-plugin-listpicker

you will notice it works fine in ios until you add cordova-plugin-googlemaps

after applying PR it works fine with both plugins

wf9a5m75:
Thank you for your cooperation. I will check it later.

wf9a5m75:
Your test code just display the list picker.
However, other people want to use the third party plugins **with maps**.

So, if I display a map, then open a list picker, here it is.

wf9a5m75:
Please try this code by yourself.

```html
<!DOCTYPE html>
<html>

<head>
  <script src="cordova.js"></script>
  <script>
    document.addEventListener('deviceready', function() {
      var mapDiv = document.getElementById('map_canvas');
      var map = plugin.google.maps.Map.getMap(mapDiv);
    });

    function listpicker() {
      try {
        window.plugins.listpicker.showPicker({
            title: "Select an option",
            items: [{
              text: "Option 1",
              value: "1"
            }, {
              text: "Option 2",
              value: "2"
            }, {
              text: "Option 3",
              value: "3"
            }],
            doneButtonLabel: "Select",
            cancelButtonLabel: "Cancel"
          },
          function(item) {
            alert("Selected: " + item);
          },
          function() {
            alert("Cancelled");
          }
        );
      } catch (e) {
        alert("Exception: " + e);
      }

      return false;
    }
  </script>
  <style type="text/css">
    html,
    body {
      width: 100%;
      height: 100%;
    }
  </style>
</head>

<body>
  <div id="map_canvas" style="width:100%;height:100%">
    <a href="#" onclick="return listpicker();">Open listpicker</a>
  </div>
</body>

</html>
```

z1ad:
Your recording is without the PR fix? The list picker is a modal view so we should not be able to interact with the map when it is open. Otherwise we cannot select an option or close the list picker.

wf9a5m75:
No. Using your PR. (Try the code I gave you)

You don't understand how manage the touch event in this plugin.
If you understand this plugin deeply, the location of your code is different.
I will change your code later.

z1ad:
OK I see what you mean. My fix only worked when the map wasnt visible, you are right.
Thank you for fixing it properly.

wf9a5m75:
I modified your code, and merge it to the multiple_maps branch.
Thank you for your PR.

datastone:
I am not being able to interact with cordova-plugin-datepicker after adding cordova-plugin-googlemaps.

PR #1880, opened 2017-11-13
add static getMyLocation function

rajeevriitm:
This creates a new static function to access `getMyLocation` without instantiating new `Map`  . It can be called with  `plugin.google.maps.Map.getMyLocation(onSuccess, onError)` .

### DIFF
--- a/www/Map.js
+++ b/www/Map.js
@@ -476,6 +476,10 @@ Map.prototype.getMyLocation = function(params, success_callback, error_callback)
             error_callback.call(self, result);
         }
     };
+    //exec undefined when called directly from plugin
+    if (!exec){
+      exec = this.exec;
+    }
     exec.call(this, successHandler, errorHandler, 'CordovaGoogleMaps', 'getMyLocation', [params], {sync: true});
 };
 Map.prototype.getFocusedBuilding = function(callback) {
--- a/www/googlemaps-cdv-plugin.js
+++ b/www/googlemaps-cdv-plugin.js
@@ -589,6 +589,7 @@ if (!cordova) {
     BaseClass: BaseClass,
     BaseArrayClass: BaseArrayClass,
     Map: {
+        getMyLocation:  Map.prototype.getMyLocation.bind({_isReady: true,exec: execCmd}),
         getMap: function(div, mapOptions) {
             var mapId;
             if (common.isDom(div)) {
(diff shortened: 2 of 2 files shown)


wf9a5m75:
Thank you for sending a PR. However not acceptable.
The geolocation should use with map.

If you want to obtain only geolocation, there are many alternatives.

rajeevriitm:
I was using `GeoLocation` plugin but it is failing on many android devices and it is one less plugin. And when tried on some devices , its faster than `Geolocation` plugin.

wf9a5m75:
I think it's almost the same implementation as the `geolocation` plugin.
(And they should tested more than my test on many devices)

wf9a5m75:
So what is the situation you want to get the location without map?

rajeevriitm:
`GeoLocation` is not working on many Xiaomi phones. Issue is open , but I but they seem to suggest its due to webview implementation but latest updates use chrome webview. And I tried their latest from dev branch but that also doesnt work (on some phones).

wf9a5m75:
Okay, then please tell me `what is the situation you want to get the location without map?`

rajeevriitm:
Its something like Tinder . Initially I need location fast no map. But I need map on another page.

wf9a5m75:
Did you consider this plugin also?
https://github.com/pmwisdom/cordova-background-geolocation-services

wf9a5m75:
I merge your PR. Thanks.

wf9a5m75:
Sorry, I revert your PR at once.
I understand your situation, but for the future plan, I will create another static class.
Just a moment please.

rajeevriitm:
Is there any problem with the implementation ?

wf9a5m75:
Your implementation itself does not have a problem.
But I will create another static class, such as `Geolocation` class, to add another methods in the feature.

wf9a5m75:
Ok, I added `Geolocation` class which is static class (internally singleton instance).
As you can see in the commit log, I will implement another methods in the future.

Please reinstall the plugin from the multiple_maps branch.
You can do like this.

```js
plugin.google.maps.Geolocation.getMyLocation(function(result) {

  var div = document.getElementById('map_canvas');
  var map = plugin.google.maps.Map.getMap(div, {
    camera: {
      'target': result.latLng,
      'zoom': 10
    }
  });
  map.on(plugin.google.maps.event.MAP_READY, function() {
    map.addMarker({
      'position': result.latLng,
      'title': result.latLng.toString()
    }, function(marker) {
      marker.showInfoWindow();
    });
  });
});
```

rajeevriitm:
That would be much better. I will check it out and get back to you.

Issue #1905, opened 2017-11-20
markerCluster MARKER_CLICK event doesn't return the clicked marker


danielcrk:
**I'm submitting a ...** 
[ ] question
[x] any problem or bug report
[ ] feature request

**If you choose 'problem or bug report', please select OS:**
[ ] **Android**
[x] **iOS**

**cordova information:** 
```
cordova-plugin-googlemaps 2.1.1 "cordova-plugin-googlemaps"
```

**Current behavior:**
The markerCluster MARKER_CLICK event callback is supposed to include 2 parameters, position and the marker. Position works properly, but marker is always undefined.

**Expected behavior:**
A reference to the clicked marker should be returned.
Also, idealy (I think), this event should be triggered when a cluster is clicked as well, and then return an array of the markers in that cluster.

**Related code, data or error log:**
```
this.map.addMarkerCluster({
  boundsDraw: false,
  maxZoomLevel: 18,
  markers,
  icons: [
    {
      min: 2,
      url: 'assets/img/geomap/map-pin-zoom-transparent.png',
      anchor: { x: 24, y: 24 },
      label: {
        color: 'white',
        fontSize: 10,
        bold: true
      },
      size: {
        width: 72,
        height: 72
      }
    }
  ]
}, (markerCluster) => {
  markerCluster.on(plugin.google.maps.event.MARKER_CLICK, (position, marker) => {
    console.log(position); // This works well
    console.log(marker); // This is always undefined
  });
});
```

Edit: The version of the plugin is 2.1.1.

danielcrk:
Also, just as a parentheses, is there a way to actually get an event when clicking a cluster?

In my case, the maxZoomLevel is set to max since the markers don't always have unique positions and I want the clustering to work when fully zoomed, but then I need to be able to actually get the click of the cluster when zoom level is 18, and somehow get all the markers in that cluster.

wf9a5m75:
This is wrong.

```
markerCluster.on(plugin.google.maps.event.MARKER_CLICK, (position, marker) => {
    console.log(position); // This works well
    console.log(marker); // This is always undefined
  });
```

Correctly

```
markerCluster.on(plugin.google.maps.event.MARKER_CLICK, (params:any[]) => {
    console.log(params);
  });
```

danielcrk:
It's the same problem, (position, marker) actually is (params[0], params[1]).

Even this just returns the location object, not the marker.
`
[...]

PR #1952, opened 2017-12-18
Add clickZoom to MarkerClusterOptions

Lingomat:
Added clickZoom boolean property to MarkerClusterOptions, similar to the ZoomOnClick property of Google's markerclustererplus. Defaults to true, so default behavior is unchanged.

If setting to false, one can use the 'cluster_click' event to decide if/when and how to zoom on cluster clicks.

### DIFF
--- a/www/Map.js
+++ b/www/Map.js
@@ -1414,7 +1414,8 @@ Map.prototype.addMarkerCluster = function(markerClusterOptions, callback) {
       "markerMap": markerMap,
       "maxZoomLevel": Math.min(markerClusterOptions.maxZoomLevel || 15, 18),
       "debug": markerClusterOptions.debug === true,
-      "boundsDraw": common.defaultTrueOption(markerClusterOptions.boundsDraw)
+      "boundsDraw": common.defaultTrueOption(markerClusterOptions.boundsDraw),
+      "clickZoom": common.defaultTrueOption(markerClusterOptions.clickZoom)
     }, exec);
     markerCluster.one("remove", function() {
       delete self.OVERLAYS[result.id];
--- a/www/Map.js
+++ b/www/Map.js
@@ -1487,7 +1487,7 @@ Map.prototype._onClusterEvent = function(eventName, markerClusterId, clusterId, 
       if (/^marker_/i.test(clusterId)) {
         // regular marker
         var marker = markerCluster.getMarkerById(clusterId);
-        if (eventName === event.MARKER_CLICK) {
+        if (eventName === event.MARKER_CLICK || eventName === event.INFO_CLICK) {
           markerCluster.trigger(eventName, position, marker);
         } else {
           if (eventName === event.INFO_OPEN) {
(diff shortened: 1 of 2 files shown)


wf9a5m75:
Why do you want to the feature? Please explain the purpose.

Lingomat:
It's so you can customise the behavior in the marker clusterer click handler. Firstly, if you have data points on top of each other, these are best left clustered (another explanation, see second comment on #1905). Zooming in further has no purpose. We are better off handling the all of the data points by, say, displaying a scrolling list on a pop up.
http://zahwa.aikuma.org is a demonstration of this using the javascript API with markerclusterer plus (you will need to  navigate to Taiwan).

Without this, the marker cluster implementation is useless for my use case.

Secondly, you can implement your own zoom. I find, for example, that cluster.getbounds() needs to be modified before passing to the maps view. Markers will be at the extreme edge of the screen and partially visible.

wf9a5m75:
>Firstly, if you have data points on top of each other, these are best left clustered 

I can not imagine the situation. Please show me the example image.

And if the current implementation is not fit for you, implementing another argorithm would be better for you (by yourself).

Lingomat:
I'll try again. Imagine ten data points with exactly the same coordinates (or very close). This is extremely common if your data are user contributions.

The algorithm is fine. Wjat you have done is 99% of the way there. This PR is a very simple change that mirrors an enhancement added in Google's markerclustererplus for exactly the same use cases. I can't see what I can do to.make this clearer. I mean I could show you a custom pop up resulting from clicking a cluster but... Does that really help?

We just want to trigger our own pop-up when a cluster is clicked on at a certain zoom level. I've also explained that the existing zoom behavior is not ideal and it'd be nice to just do your own to zoom if you want. This requires stopping the current click to zoom behavior, that's all.

Lingomat:
clusterpopup

An example using Javascript API on desktop with Google's markerclustererplus. This is a highly zoomed map, all seven markers are recipes by the same person, so are very close together. In this app, I simply disable click to zoom and handled it manually. When the zoom level is at a certain point, clicking on the cluster does not zoom, instead we trigger a pop up to show all of the data.

wf9a5m75:
You can modify on forked repository. Why do you need to send this PR? Smart phone has small display than desktop. So I can't agree such a popup is good for everyone

Lingomat:
Okay this is getting ridiculous. The example shot is from a desktop app, on mobile my forked repo obviously does something suitable for mobile, that is irrelevant.

I've also explained to you that the current zoom is not great (am I'm being polite), and it's totally reasonable for us to implement the zoom ourselves. Jesus for the sake of six lines that does nothing but provide an *option* allow us to bypass some really inflexible and sub optimal code... An option is not *everyone*, and we PR because we hope to maintain parity with upstream and benefit from ongoing maintenance and bug fixes. 

Anyway forget it, I'll just join the other 600 forks until I can replace this whole thing... sheesh.

PR #1996, opened 2018-01-15
Use console.error instead of displaying an alert

sburnicki:
When having a code base for an app that can run both in the browser and on android, it might not be that easy to exclude this plugin while maintaining an easy codebase/build system.

So if the plugin fails to initialize, it might still make sense to use the application without showing errors directly to the user. Therefore I suggest to replace the very intruding alert, by a more silent `console.error`, so the UX is not directly affected by the error.

### DIFF
--- a/www/googlemaps-cdv-plugin.js
+++ b/www/googlemaps-cdv-plugin.js
@@ -675,10 +675,10 @@ if (!cordova) {
           if (/Android/i.test(window.navigator.userAgent)) {
               //------------------------------------------------------------------------
               // If Google Maps Android API v2 is not available,
-              // display the warning alert.
+              // display the error.
               //------------------------------------------------------------------------
               cordova.exec(null, function(message) {
-                  alert(message);
+                  console.error(message);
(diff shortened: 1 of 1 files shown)


wf9a5m75:
The most js developers who ask to bug reports to me do not check even console log.
And Safari debugger does not obtain the error logs which are outputted before open the debugger are not oppened. In that case, the mobile developers are not able to notice the error message.
So alert is the best way.

sburnicki:
In this case, you're making it really hard for projects that support multiple platforms, including browser. Cordova and Ionic infrastructure are both not that good that they make it easy to exclude this plugin easily for browser platform build without causing difficulties in code.

JS developers can easily check the console. And even with safari debugger, they just need to reload the page with Cmd+R. Devs can easily do that, so is this really worth intruding user experience?

sburnicki:
Also, you could just add checking the console log to the issue template, to make sure developers do this before reporting.

wf9a5m75:
Use the `multiple_maps` branch. You don't see the alert on browser platform.

PR #2118, opened 2018-03-01
Add g and path tags to ignored tags list.

leandroz:


wf9a5m75:
What the tags `g` and `path`?

leandroz:
@wf9a5m75 please check the referenced issue, `className`is not a string on those 2 tags.

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/path

wf9a5m75:
Since I inserted different code, I close this PR. But thank you for your PR anyway.

PR #2174, opened 2018-03-23
Fix Dutch language filename on iOS

d3r1w:
First thanks for all your work!
I encountered a small bug while installing the multiple_maps branch for an iOS app im developing which results in a file not found error of the Danish language files. The expected filename is *pgm_Localizable_da.json* but was *pgm_Localizable_da.strings*. After I renamed the file the installation worked as expected.

I hope the change is to your liking and please let me know if you need some further adjustments.


wf9a5m75:
Thanks!

PR #2189, opened 2018-04-01
Added improvements to Tile Overlay

Kraxxis:
Added the ability to return a Promise from the getTile TileOverlay
handler.
Added "invalidate" method to Tile overlay to allow invocation of the GMS
"clearTileCache" method from JS.
Added ability to return a base64 encoded image from getTile in JS.

---

My project requires the ability to dynamically generate Tiles at runtime.  These changes would allow my project to do this.

Note I'm having a tough time attempting to test these changes.  I will be able to do more once the holiday is over.  I have a demo project developed here in an attempt to test these changes: https://github.com/Kraxxis/cordova-googlemaps-base64-example.

I also left the Android implementation of base64 handling absent; I am unsure how to do this in Java land, and it looks like there is some advanced caching going on that is not present in iOS.  Any advice you have would be helpful and appreciated.

### DIFF
--- a/src/android/plugin/google/maps/PluginTileOverlay.java
+++ b/src/android/plugin/google/maps/PluginTileOverlay.java
@@ -199,6 +199,23 @@ public class PluginTileOverlay extends MyPlugin implements MyPluginInterface {
   }
 
   /**
+   * Clears the tile cache
+   * @param args
+   * @param callbackContext
+   * @throws JSONException
+   */
+  public void invalidate(final JSONArray args, final CallbackContext callbackContext) throws JSONException {
+      String id = args.getString(0);
+      TileOverlay tileOverlay = (TileOverlay)pluginMap.objects.get(id);
+      if (tileOverlay == null) {
+          callbackContext.error("TileLayer(" + id + ") not found");
+      } else {
+          tileOverlay.clearTileCache();
+          callbackContext.success();
+      }
+  }
+
+  /**
    * Set fadeIn for the object
    * @param args
    * @param callbackContext
--- a/src/android/plugin/google/maps/PluginTileProvider.java
+++ b/src/android/plugin/google/maps/PluginTileProvider.java
@@ -176,7 +176,14 @@ public class PluginTileProvider implements TileProvider  {
 
     try {
       InputStream inputStream = null;
-      if (urlStr.startsWith("http://") || urlStr.startsWith("https://")) {
+      if (urlStr.startsWith("data:image") {
+        //-------------------------------
+        // load image from base64 string
+        //-------------------------------
+        
+        // not yet supported
+        return null;
+      } else if (urlStr.startsWith("http://") || urlStr.startsWith("https://")) {
         //-------------------------------
         // load image from the internet
         //-------------------------------
--- a/src/ios/GoogleMaps/PluginTileOverlay.h
+++ b/src/ios/GoogleMaps/PluginTileOverlay.h
@@ -19,6 +19,7 @@
 -(void)create:(CDVInvokedUrlCommand*)command;
 -(void)setVisible:(CDVInvokedUrlCommand *)command;
 -(void)remove:(CDVInvokedUrlCommand *)command;
+-(void)invalidate:(CDVInvokedUrlCommand *)command;
 -(void)setZIndex:(CDVInvokedUrlCommand *)command;
 -(void)setFadeIn:(CDVInvokedUrlCommand *)command;
 -(void)setOpacity:(CDVInvokedUrlCommand *)command;
--- a/src/ios/GoogleMaps/PluginTileOverlay.m
+++ b/src/ios/GoogleMaps/PluginTileOverlay.m
@@ -218,7 +218,19 @@
   }];
 
 }
+-(void)invalidate:(CDVInvokedUrlCommand*)command 
+{
+    [self.executeQueue addOperationWithBlock:^{
+      NSString *tileLayerKey = [command.arguments objectAtIndex:0];
+      dispatch_async(dispatch_get_main_queue(), ^{
+          GMSTileLayer *layer = (GMSTileLayer *)[self.mapCtrl.objects objectForKey:tileLayerKey];
+          [layer clearTileCache];
+      });
 
+      CDVPluginResult* pluginResult = [CDVPluginResult resultWithStatus:CDVCommandStatus_OK];
+      [self.commandDelegate sendPluginResult:pluginResult callbackId:command.callbackId];
+  }];
+}
 
 /**
  * Set z-index
--- a/src/ios/GoogleMaps/PluginTileProvider.m
+++ b/src/ios/GoogleMaps/PluginTileProvider.m
@@ -116,6 +116,15 @@ NSDictionary *debugAttributes;
      }
   }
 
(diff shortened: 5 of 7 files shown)


wf9a5m75:
In your example, you can generate base64 encoded image without promise.

Kraxxis:
True. However, in my application, which is not shown here, i need to query external services to produce the data needed to generate the image based on the x, y, and zoom values at runtime.

The purpose of the example was just to exercise the functionality i added.

Kraxxis:
I've expanded my example with a "delay" acting as an external query.

wf9a5m75:
I don't want to allow `external action` for the `getTile()` callback, because end user don't want to wait long time.
Developer should implement additional query in server side.

wf9a5m75:
So I have a question, in order to display the custom tile, do you want to wait 10 or more longer seconds ?
Even just a few seconds, you would feel kind of stress.

Kraxxis:
> Developer should implement additional query in server side.

I cannot do this; i do not control the server side.

> I don't want to allow external action for the getTile() callback, because end user don't want to wait long time.

Thats still a problem if `getTile` returns an `http` url.  This pattern allows my JavaScript to process the result first.

> So I have a question, in order to display the custom tile, do you want to wait 10 or more longer seconds ?

No, i think something longer than 10 seconds would not be good.  Realistic latency would be 1-2 seconds _at most_.

Perhaps the timeout before native giving up could be configurable? 

Longer queries and delays would be the responsibility of my application to handle.  If I am concerned about latency, my application would add a timeout and if it fails would return a null in the promise result.

wf9a5m75:
Actually native side has 10 seconds timeout.

----

However, you don't recognize what the problem is.
**The `mobile` network always slow than PC.  This is extremely important.**

For example, even just 100kb for one tile image, the app takes several seconds on poor network.
So you want to query to your server (for example), it takes a few seconds, then the plugin requests the tile image with a few seconds.
Thus, it takes from 5 to 10 seconds on poor network.

image1

And the `getTile()` callback is executed on every tile image request.
So the maps plugin requests 8 tile images, your app requests 16 times http requests.

However if you return the url with query information, your apps requests only 8 times.
image2

You need to know `http request` needs lots of battery usage.
**Thus, mobile app has to save the number of http requests as much as possible.**

That's why I can not accept your pull request.

-----------

>i do not control the server side.

The `Google Cloud Functions` is one of good choices for you.
You can create server side code without having a server.
Automatic scale, easy to use it (node.js), and cheap price.

wf9a5m75:
FYI. Google Cloud Functions
https://cloud.google.com/functions/

wf9a5m75:
Regarding of the `invalidate` method, when do you want to execute it?
Does your app monitor the memory usage?

`Cache` means `preparing the result for next time of the same request`.
As I described above, **mobile app has to save the number of http request as much as possible**.
Do you decide exactly `when`?

In the native side, the plugin executes the `clearTileCache` method when the tile layer is removed.
I think this is enough, because the native Google Maps APIs also control the cache. If the APIs decide the cache is too much, the cache data is removed internally.

But you want to execute the method. Could you explain `when` do you want to execute the method from JS side?

wf9a5m75:
Please don't misunderstand, I have not reject you.
I just only discuss the problems in logical.

I really appreciate for your action to create a pull request itself.

Kraxxis:
Thank you taking the time to talk this through.

From your graphic, i don't think you understand my use case. I will try to explain.

My application needs to display real time data on the map.  Right now, my application displays this data using MapPolygons.  However, I have too much data, and GoogleMaps cannot handle lots of polygons on the map.  This data does not need to be interacted with (no touches), so keeping the polygon information is too much.

Instead, i want to take the 1000s of MapPolygons from the server and transform the data into an image to display on the Map.

1000s of MapPolygons ---> 1 Tile Image should help me quite a bit with performance.

Right Now:
```
(( Native )) ------------------- (( JavaScript )) --------------------- (( Server ))
      |                                 |                                      |
                                        | --- HTTP Get All JSON Data --> |
                                        | <---------  JSON Data -------- |
      |               << Create Map Polygons from JSON Data >> 
      | <------ Add Map Polygons ------ |
```
What I need to do, and what these changes allow:
```
(( Native )) ------------------- (( JavaScript )) --------------------- (( Server ))
      |                                 |                                      |
      | ---------- getTile -----------> |                                      |
      |                                 | ---- HTTP Get JSON only for tile --> |
      |                                 | <------------ JSON Data ------------ |
      |                  << Create Image from JSON Data >>                     |
      | <-- base64 getTile Response --- |                                      |
```
So there is still only one HTTP request per tile.

I want to use TileLayer instead of GroundOverlay because:
1.  My application doesn't know where to query, I need to display what is being shown, not all data
2.  By using Tiles I reduce the size of the HTTP Get which makes the requests faster.

My application sometimes gets notified the data on the server has changed, and i need to recreate the images.  `invalidate` would let me start the process over and allow me to clear the stale data.

I want to emphasize, this data is NOT static, but dynamic.  The data changes, and i need to update periodically, perhaps once a minute.

Kraxxis:
> Cache means preparing the result for next time of the same request

I agree.  I plan on adding a Base64 Image Cache in JS.

However sometimes the server decides the data changes. I need to tell GoogleMaps to clear the images on the screen so I can remake them.

wf9a5m75:
I think you create your app with native code is better.

As I said, the `getTile()` is executed for every tiles. In order to dispaly 8 tiles, 8 http requests.
Not only one time.

And generating dynamic images with JavaScript use too much CPU. Because the browser executes JavaScript on main thread.
This is really bad performance.

Thus, you should create your app with native code instead of this plugin.

wf9a5m75:
If you want to use this plugin with your purpose anyway, I recommend you create a large hidden canvas, and draw a bitmap with your realtime data.
Then you just splice it and return it as tile with base64 image.

This is the most efficient way. Only one http request, but you can reply the image in synchronized.

This is just idea, I haven't tested it, but it should be fine.

Kraxxis:
> I think you create your app with native code is better

I understand where you are coming from, but I cannot do this; this is a piece of a larger app, and there are other features which require the Cordova architecture.

My team has done internal testing and have shown drawing on a hidden canvas, like in my example, is very fast and efficient, even in JavaScript. The largest hit would be creation of the base64 string.

> This is the most efficient way. Only one http request, but you can reply the image in synchronized.

Yes, creating the image synchronously is the plan, but I need `getTile` to allow asynchronous flow for requesting the external data.  The request must be asynchronous.

Are there any changes you would like to see in this PR? This is what I thought would be the minimum and simplest changes needed in this project to support the implementation I need.

I understand if you do not want to incorporate these changes. My project could move forward with a fork, but it would be much cleaner to have these changes in mainline, and I thought this project would benefit from these simple improvements.  My goal was simply to expose more of the GMS API (Asynchronous tile creation and `clearTileCache`) from this plugin to JS, not to add anything large or unwieldy.

I will be testing these changes in earnest tomorrow.

wf9a5m75:
You have enough knowledge for creating good application, but nost other people are not.
I recommend moving forward with your forked version is good at this time.

wf9a5m75:
Although I don't merge your PR at this time, I will provide another class that idea of big canvas in the future.

wf9a5m75:
Thank you for discussing.

PR #2193, opened 2018-04-03
Improvements to coordinate precision

tsbgarner:
Coordinate precision was being lost in iOS (all cases) and Android (kml overlay) when more than 6 decimal places were required due to using floats instead of doubles. Due to this, rendered coordinates could be potentially out by a meter.


wf9a5m75:
Great! Thank you for your help!

PR #2289, opened 2018-05-26
fix error parentElement of null

lenguyentrang:
fix error parentElement of null

### DIFF
--- a/www/CordovaGoogleMaps.js
+++ b/www/CordovaGoogleMaps.js
@@ -103,12 +103,16 @@ function CordovaGoogleMaps(execCmd) {
       }
     });
   });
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Could you explain this situation?

lenguyentrang:
when we have require CordovaGoogleMaps.js but the body page not yet for ready load =>document.body will be null => we have error for "null.parent".   
PS: I try to change on my local it run OK. But I get issue the map can be display but can not dragable, click,...

wf9a5m75:
Thank you for sending your PR. Since your PR just declines for the case, I added different code. It works perfectly. So I close your PR.
But thank you for sending your PR anyway.

PR #2341, opened 2018-07-09
[iOS] Fixed bug in "getMyLocation" with last location result

AKosmachyov:
Fixed bug with wrong coordinates of the current location on iOS.

Old behavior:

screen shot 2018-07-09 at 5 57 53 pm


wf9a5m75:
Oh, thank you for finding the bug.

PR #2388, opened 2018-08-09
another compilation fix

csabbey:
I fixed this problem in another file (which you merged), but it popped up here, too. I forked the project and I'm running from that, so I'm pretty sure this is the last one.


wf9a5m75:
The `v2.3.11` covers this PR. But thank you for your PR anyway.

PR #2400, opened 2018-08-21
Update Cluster.js

anakinwq:
Failing to have a method "getItemVisibleLength()", it would be good in the logic of not raised in the count of the markerCluster the sum of markers visible or not visible.

What do you think ?

### DIFF
--- a/www/Cluster.js
+++ b/www/Cluster.js
@@ -112,7 +112,14 @@ Cluster.prototype.remove = function() {
   });
 };
 Cluster.prototype.getItemLength = function() {
-  return this._markerOptsArray.length;
+  //return this._markerOptsArray.length;
+   var inc = 0;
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Most current user do not have `visible` field. Did you consider what would be occurred?

wf9a5m75:
Instead of implementing particular field filtering in the main code, how about external filter function?
For example,

```js
var markerCluster = map.addMarkerCluster({
  filter: function(marker) {
    return marker.visible === true
  }
});
```

wf9a5m75:
This PR is covered in https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/ecf48a22a0a80e4aa227829543daf9f9142a69c7 by different way.

From the next version, the marker cluster ignores invisible markers.

Thank you for your PR anyway.

Issue #2409, opened 2018-08-30
Reopen #1919 & #2291: Bottom part of ionic tabs is not clickable


KiwiKilian:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [x] any problem or bug report

**OS:**  (check one with "x")
- [ ] **Android**
- [x] **iOS**
- [ ] **Browser**

**cordova information:**  (run `$> cordova plugin list`) 

```
com.googlemaps.ios 2.7.0 "Google Maps SDK for iOS"
cordova-plugin-device 2.0.2 "Device"
cordova-plugin-googlemaps 2.4.1 "cordova-plugin-googlemaps"
cordova-plugin-ionic-keyboard 2.1.2 "cordova-plugin-ionic-keyboard"
cordova-plugin-ionic-webview 2.1.0 "cordova-plugin-ionic-webview"
cordova-plugin-splashscreen 5.0.2 "Splashscreen"
cordova-plugin-statusbar 2.4.2 "StatusBar"
cordova-plugin-whitelist 1.3.3 "Whitelist"
```

**If you use `@ionic-native/google-maps`, please tell the package.json (only `@ionic-native/core` and `@ionic-native/google-maps` are fine mostly)**
```
@ionic-native/core : 4.12.0
@ionic-native/google-maps : 4.12.0
```

**Current behavior:**
Bottom part of Ionic tabs are not clickable when Cordova Google Maps is installed.
Same issue as described in #1919 and #2291, seems still not fixed. When Cordova Google Maps is installed the tabs are only clickable in the upper two thirds, as it can be seen in the following video.

<img src="https://user-images.githubusercontent.com/18724559/44875194-be754700-ac9d-11e8-8fdc-56817e8c8eb8.gif" width="250">

**Expected behavior:**
Bottom part of Ionic tabs clickable. For testing I completely removed the Cordova Google Maps Plugin. Afterwards the tabs are clickable everywhere, just as expected. See the second video below.

<img src="https://user-images.githubusercontent.com/18724559/44875195-be754700-ac9d-11e8-900b-474ea35995ed.gif" width="250">

**Repository:**
[Repository with non working example](https://github.com/KiwiKilian/ionic-google-maps-debug)

wf9a5m75:
I just built and run your repo.
![2409](https://user-images.githubusercontent.com/167831/44878436-bac0d080-ac5b-11e8-92b6-4221f8cbd78f.gif)

[...]

Issue #2055, opened 2018-02-07
touch problem with cordova-plugin-statusbar


shrijan:
**I'm submitting a ...**  (check one with "x")
- [x] question
- [x] any problem or bug report
- [ ] feature request

**If you choose 'problem or bug report', please select OS:**  (check one with "x")
- [ ] **Android**
- [x] **iOS**

**cordova information:**  (run `$> cordova plugin list`) 

```
com.googlemaps.ios 2.5.0 "Google Maps SDK for iOS"
cordova-plugin-actionsheet 2.3.3 "ActionSheet"
cordova-plugin-badge 0.8.5 "Badge"
cordova-plugin-calendar 4.6.0 "Calendar"
cordova-plugin-camera 3.0.0 "Camera"
cordova-plugin-cocoapod-support 1.3.0 "Cordova CocoaPods Dependency Support"
cordova-plugin-compat 1.2.0 "Compat"
cordova-plugin-console 1.1.0 "Console"
cordova-plugin-device 1.1.7 "Device"
cordova-plugin-dialogs 1.3.4 "Notification"
cordova-plugin-file 4.3.3 "File"
cordova-plugin-file-transfer 1.6.3 "File Transfer"
cordova-plugin-firebase 0.1.25 "Google Firebase Plugin"
cordova-plugin-geolocation 3.0.0 "Geolocation"
cordova-plugin-googlemaps 2.2.3 "cordova-plugin-googlemaps"
cordova-plugin-inappbrowser 1.7.2 "InAppBrowser"
cordova-plugin-keyboard 1.2.0 "Keyboard"
cordova-plugin-local-notification 0.9.0-beta.1 "LocalNotification"
cordova-plugin-nativeclicksound 0.0.3 "Native Click Sound"
cordova-plugin-splashscreen 4.1.0 "Splashscreen"
cordova-plugin-statusbar 2.4.1 "StatusBar"
cordova-plugin-tts 0.2.3 "TTS"
cordova-plugin-whitelist 1.3.3 "Whitelist"
cordova-plugin-x-socialsharing 5.2.1 "SocialSharing"
cordova.plugins.diagnostic 3.6.7 "Diagnostic"
es6-promise-plugin 4.1.0 "Promise"
ionic-plugin-keyboard 2.2.1 "Keyboard"
uk.co.workingedge.phonegap.plugin.launchnavigator 4.1.0 "Launch Navigator"
```

[...]

PR #2411, opened 2018-09-02
Fix bottom area clickable #2409

KiwiKilian:
This fixes the problem when using `corodova-plugin-googlemaps` together with `cordova-plugin-statusbar` to make the bottom 20px clickable (#2409).
When running the hitTest on the subviews the height of the statusbar has to be accounted for. I tested my changes with and without the `cordova-plugin-statusbar` beeing installed, both ways worked fine.

### DIFF
--- a/src/ios/GoogleMaps/MyPluginLayer.m
+++ b/src/ios/GoogleMaps/MyPluginLayer.m
@@ -330,6 +330,8 @@
   // e.g. PhoneGap-Plugin-ListPicker, etc
   UIView *subview;
   NSArray *subviews = [self.webView.superview subviews];
+  CGRect statusBarFrame = [UIApplication sharedApplication].statusBarFrame;
+  CGPoint subviewPoint = CGPointMake(browserClickPoint.x, browserClickPoint.y - statusBarFrame.size.height);
   for (int i = ((int)[subviews count] - 1); i >= 0; i--) {
     subview = [subviews objectAtIndex: i];
     //NSLog(@"--->subview[%d] = %@", i, subview);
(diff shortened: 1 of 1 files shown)


wf9a5m75:
You can find in whcich commit from blame history of GitHub. I'm in a mountain now for hiking. If you can't find in which commit, please wait until I'm ready.

KiwiKilian:
No rush, really thankful for any of your time improving this plugin!  I'm just interested in getting this problem fixed and don't want to introduce any new bugs.

I didn't just copy the code. I only applied the statusbar difference for the subviews while the initial fix, which you reverted over time, changed a lot more.
This is the debug output for one of the subviews:
```
--->subview[2] = <UIScrollView: 0x7fa231064000; frame = (0 0; 375 667); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x60c000255cc0>; layer = <CALayer: 0x608000429760>; contentOffset: {0, 647}; contentSize: {375, 1334}; adjustedContentInset: {20, 0, 0, 0}>
```
The `contentOffset` and `adjustedContentInset` led me to the conclusion, that the hitTest on those subviews need to subtract the statusbar height because any hit below 647px wouldn't be inside of those subviews. Also without the fix those klicks aren't going anywhere, the debug shows it's not resulting in `in map` and also not `in browser`.

 Here are the related commits:
-  Initial attempt to fix the bottom clickable problem
   - https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/d41f4362df341161a5119a46b6d979e9c2aa2e3b
- Due to #2055 the fix was reverted over multiple commits
  - https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/39131e5e5f4e4fe7a642fe512435a5ccd3c7aef6
  - https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/46ea90d83ac2262deb1ab55f4db44ce5c83cb97c
  - https://github.com/mapsplugin/cordova-plugin-googlemaps/commit/4a2e8d501a17bd0eecd735877bfca3c93227b5bc

I tested the example repo mentioned in #2055, everything seems to be working fine with my solution.

PR #2422, opened 2018-09-17
(iOS) Fixed issue where plugin could appear behind other plugins

thes3m:
This bugfix should solve issues on iOS where another plugin might position itself above map view in view hierarchy and therefore hide map view. The plugin simply makes sure that the map view is right below the webview each time it is presented to prevent this issue. I guess this issue can occur in combination with other plugins that might use native views such as camera etc.


wf9a5m75:
I tested the maps plugin with `cordova-plugin-cszbar`, but I could not reproduce the issue situation.
However it could be possible.
So I merge your PR.
I really appreciate for your contribution.

thes3m:
Nice. You are welcome.

PR #2425, opened 2018-09-18
plugin.xml: Use github: prefix for dependency

jayvdb:
`https://...` links are for tarballs, according to the spec.  While `npm` does detect it is a git link, other package managers do not, such as https://github.com/pnpm/pnpm/issues/1376 .

Either `git+` or `github:` are required for git links.  Most people now use a package manager version which supports `github:`

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -476,7 +476,7 @@
       </array>
     </config-file>
 
-    <dependency id="com.googlemaps.ios" url="https://github.com/mapsplugin/cordova-plugin-googlemaps-sdk.git#2.7.0"/>
+    <dependency id="com.googlemaps.ios" url="github:mapsplugin/cordova-plugin-googlemaps-sdk#2.7.0"/>
 
     <hook type="before_platform_add" src="src/ios/check_sdk_version.js"/>
     <hook type="before_plugin_install" src="src/ios/check_sdk_version.js"/>


wf9a5m75:
>https://... links are for tarballs, according to the spec.

Please refer the document.

There is no problem with current style so far. In order to accept your PR, I need to test it several situations, such as old npm.

wf9a5m75:
>Most people now use a package manager version which supports github:

Please provide which npm version support this. The word `Most people` is not reliable word if there is ensure data.

wf9a5m75:
Regarding of the npm document, `github:` protocol executes `git clone` action, however some people (mostly windows super beginners) report `git clone` problem.

Regarding of your comment, `https:` is for `tarball`.
So, let's use tarball `https://github.com/mapsplugin/cordova-plugin-googlemaps-sdk/tarball/2.7.0`.

wf9a5m75:
v2.4.4 specifies the url to tarball.

https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/3e8879984d9224d4803d9655a1232c3deea174b3/plugin.xml#L479

If pnpm implements correctly as you describe, this should work, shouldn't it?
Npm works correctly at least.

jayvdb:
Great, thanks.  2.4.4 works for me, with pnpm, and tarball links are better.

Wy not do registry releases of the sdk package?  That would be even better.  A bit more work though, and doesnt really effect me, so no drama if you are skipping it for that reason.  I'm mostly curious.

wf9a5m75:
> Wy not do registry releases of the sdk package? 

Because it's too heavy.

wf9a5m75:
@jayvdb `https://... links are for tarballs, according to the spec. `
Please tell me where you find this document ** immediately**.

At least, this is not correct for all the people.
https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/2427

PR #2432, opened 2018-09-21
Feature test for viewport-fit=cover

chadevans:
Current functionality checks user agent for iOS 11 or higher. If `true`, adds `viewport-fit=cover` to the `viewport` meta tag. This allows use of safe areas, for example, the "notch" at the top of iPhone X.

Google now [supports this feature](https://blog.chromium.org/2018/08/chrome-69-beta-av1-video-decoder-css.html) in Chrome 69, opening the door for Android phones to use Display Cutouts.

I propose this plugin should test the browser for support of the CSS variable `env()` which delivers the pixel value of safe spaces, like this:

#box {
  margin-top: env(safe-area-inset-top);
  margin-left: env(safe-area-inset-left);
  margin-bottom: env(safe-area-inset-bottom);
  margin-right: env(safe-area-inset-right);
}

The included test creates a div with a `margin-top` set to `-99px` followed by `margin-top:env(safe-area-inset-top)`.

If the browser does not support `env()`, then `margin-top:env(safe-area-inset-top)` is ignored and the calculated value remains `-99px`. In this case, `viewport-fit=cover` is not added to the `viewport` meta tag.

If the browser does support `env()`, the calculated value will be something other than `-99px`, either `0px` or a specific value associated with that phone. In this case, `viewport-fit=cover` is added to the `viewport` meta tag, allowing use of `env()` for display cutouts.

### DIFF
--- a/www/pluginInit.js
+++ b/www/pluginInit.js
@@ -63,15 +63,22 @@ function pluginInit() {
 
   var viewportTagContent = 'width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no';
 
-  // Detect if iOS device
-  if (/(iPhone|iPod|iPad)/i.test(window.navigator.userAgent)) {
-    // Get iOS major version
-    var iosVersion = parseInt((window.navigator.userAgent).match(/OS (\d+)_(\d+)_?(\d+)? like Mac OS X/i)[1]);
-    // Detect if device is running >iOS 11
-    // iOS 11's UIWebView and WKWebView changes the viewport behaviour to render viewport without the status bar. Need to override with "viewport-fit: cover" to include the status bar.
-    if (iosVersion >= 11) {
-      viewportTagContent += ', viewport-fit=cover';
-    }
+  //
+  // Detect support for CSS env() variable
+  //
+  var envTestDiv = '<div id="envTest" style="margin-top:-99px;margin-top:env(safe-area-inset-top);position:absolute;z-index:-1;"></div>';
+
+  document.body.insertAdjacentHTML('afterbegin', envTestDiv);
+
+  var testElement = document.getElementById('envTest');
+  var computedStyles = window.getComputedStyle(testElement);
+  var testResult = computedStyles.getPropertyValue('margin-top');
+
+  document.body.removeChild(testElement);
+
+  // if browser supports env(), returns a pixel value as string, even if 0px
+  if (testResult != '-99px') {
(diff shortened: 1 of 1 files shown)


chadevans:
I have done testing, my results were as expected, otherwise would not have submitted.

Testing in an Ionic app, Chrome 69 on Android 8 (and Android System Webview 69 on Android 5 & 6) returned `0px`. Doesn't matter the OS on Android, it matters the Chrome or Android System Webview version. Version 69+ includes `env()` support, otherwise would return `-99px`.

Ran the same Ionic app using Chrome 66 on an Android 9 emulator, it returned `-99px`, as `env()` is not supported pre-69.

In iOS testing, this also worked as expected. iPhone X returned `44px` as it not only supports `env()`, it uses the feature for the notch cut-out.

Using an iPhone 6s emulator running iOS 10, I got `-99px`, as `env()` support was added in iOS 11.

iPhone 8 Plus running iOS 11 returned `0px`. Expected.

wf9a5m75:
Thank you for creating clever solution. I really appreciate for your help!

chadevans:
Thank you for the fantastic plugin. Love it!

chadevans:
> @chadevans I get "Uncaught TypeError: Cannot call method 'insertAdjacentHTML' of null" on android API 19.

Hm, error appears to say there is no `document.body` yet... please create new issue, I'll do some testing.

chadevans:
@cpamp I am using Ionic which likely already has a DOM when this plugin runs... no easy way to test. Try this solution inside pluginInit.js to control when the test div is created: https://stackoverflow.com/questions/26324624/wait-for-document-body-existence

cpamp:
There's 2 existing issues already (closed). I think the issue is related to upgrading. I'll try to take a closer look over the next few days. Everyone who reported the error was on 2.4.4 and upgraded to 2.4.6 but I did not get the error when starting with a blank project.

chadevans:
@cpamp if you are in a cordova project, not Ionic, could you check the DOM and see if the script loads in the `<head>` which is before `<body>`.

cpamp:
It's almost certainly a race condition.

I am in Ionic.

In this project:

```
 // The pluginInit.js must execute before loading HTML is completed.
  require("./pluginInit")();
```

Body is null because the page hasn't loaded. Wrapping that entire viewport section in a `setTimeout` seems to work. I'll look into it and hopefully PR this weekend. I have this issue open: #2445 

Question though, why is this plugin adding `viewport-fit=cover` anyways?

PR #2446, opened 2018-10-03
Android transparent webview fix

thes3m:
Prevented a bug that can occur when using google maps in conjunction with other plugins that also modify webview background color. The plugin did not apply background color each time it was presented, therefore it was possible that some other plugin could change the background color and make the webview opaque.


thes3m:
Sure, Here you go.

wf9a5m75:
Thank you for your contribution ;)

thes3m:
You are welcome. ;)

PR #2466, opened 2018-10-15
Fix crash for POI with quote in name

supa-creation:


wf9a5m75:
Thank you for your contribution

PR #2483, opened 2018-10-25
Implement testing for project

adamduren:
Added test scaffolding so that we could add tests to this project going forward.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 www/Promise.js
+*.spec.js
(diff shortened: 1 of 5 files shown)


.eslintignore
```diff
@@ -1 +1,2 @@
 www/Promise.js
+*.spec.js
```
adamduren: Ignore the `spec` files as we can use fancy things such as const, let, and arrow functions in jest tests.

jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
adamduren: Similar to what we saw in the ionic-native version. We could also implement TypeScript for just the specs to get type helping but I'll only add it if that's the direction you want to take this.

wf9a5m75:
I really appreciate for your great work!

PR #2500, opened 2018-11-12
Add a custom resolver so that we can require our own project files

adamduren:
The following PR makes it to where files that `require` files under the plugin namespace can be resolved. 

```
var event = require('cordova-plugin-googlemaps.event'),
  BaseClass = require('cordova-plugin-googlemaps.BaseClass'),
  LatLng = require('cordova-plugin-googlemaps.LatLng'),
  MapTypeId = require('cordova-plugin-googlemaps.MapTypeId');
```

Will now be correctly resolved to the correct location. Since Cordova implements it's own `require` function we need to customize jests when importing something from our own plugin.

### DIFF
--- a/jest-cordova-resolver.js
+++ b/jest-cordova-resolver.js
@@ -1,0 +1,22 @@
+var Resolver = require('jest-resolve');
+var path = require('path');
+var fs = require('fs');
+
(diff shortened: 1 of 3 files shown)


wf9a5m75:
Thank you. I faced the same problem on another project, then I resolved different way. But your solution is smart.

PR #2502, opened 2018-11-12
Implement demo of using JSDoc to support type checking js

adamduren:
@wf9a5m75 I am curious to get your thoughts on adding the type checking to the project without having to convert the entire project to typescript. Would this be something that you are in favor of?

Currently there are types in the `ionic-native` repo. It seems like it would be beneficial to be able to reference them in this project as well to help keep the types in sync as well as providing type checking to this project which will help reduce chance of errors.

This PR is to demo how it could be achieved. We could export the types from this project using the `package.json` `types | typings` field and then the `ionic-native` plugin could list this as a `devDependency`. Another option would be to put the types in their own repo in which both this repo and the ionic-native one would depend on but I am leaning towards moving them into this project in order to reduce complexity.

It would be an effort I can help with over time but as we move types over we can list files in the `includes` field of `tsconfig` and that will provide type checking moving forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     }
   ],
   "scripts": {
-    "test": "jest && npm run eslint",
+    "check-types": "tsc",
+    "test": "npm run check-types && npm run jest && npm run eslint",
     "jest": "jest",
     "test:watch": "jest --watch",
     "eslint": "node node_modules/eslint/bin/eslint www src"
@@ -55,8 +56,10 @@
   },
   "homepage": "https://github.com/mapsplugin/cordova-plugin-googlemaps",
   "devDependencies": {
+    "@types/googlemaps": "^3.30.16",
     "cordova-js": "^4.2.4",
     "eslint": "^5.7.0",
-    "jest": "^23.6.0"
+    "jest": "^23.6.0",
+    "typescript": "^3.1.6"
   }
 }
--- a/src/browser/PluginGeocoder.js
+++ b/src/browser/PluginGeocoder.js
@@ -1,5 +1,6 @@
 
 
+// @ts-ignore
 var BaseArrayClass = require('cordova-plugin-googlemaps.BaseArrayClass');
 
 var geocoder = null;
@@ -146,6 +147,12 @@ QUEUE.on('next', function() {
 });
 
 module.exports = {
+  /**
+   *
+   * @param {(result: GeocoderResult[]) => void} onSuccess
+   * @param {(err: string) => void} onError
+   * @param {[GeocoderRequest]} args
+   */
   'geocode': function(onSuccess, onError, args) {
     var request = args[0];
     var geocoderRequest = {};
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -43,7 +43,8 @@
     }
   ],
   "scripts": {
-    "test": "jest && npm run eslint",
+    "check-types": "tsc",
+    "test": "npm run check-types && npm run jest && npm run eslint",
```
adamduren: Add type checking as part of the test process.

package.json
```diff
@@ -55,8 +56,10 @@
   },
   "homepage": "https://github.com/mapsplugin/cordova-plugin-googlemaps",
   "devDependencies": {
+    "@types/googlemaps": "^3.30.16",
```
adamduren: This will make sure we are passing the appropriate params to google maps api.

src/browser/PluginGeocoder.js
```diff
@@ -1,5 +1,6 @@
 
 
+// @ts-ignore
```
adamduren: This is needed because of cordova custom module resolution

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "allowJs": true,
```
adamduren: `allowJs` and `checkJs` are what enable type checking. `noEmit` is added because we don't actually want to compile the files.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "checkJs": true,
+    "noEmit": true,
+    "esModuleInterop": true
+  },
+  "include": [
+    "src/browser/PluginGeocoder.js",
```
adamduren: Add sources as we go to incrementally type the project over time.

adamduren:
Just for an example here is the output for `npm run test` catching a typo I made for demonstration purposes.

```zsh
➜  cordova-plugin-googlemaps git:(discuss-embeding-type-info) ✗ npm run check-types

> cordova-plugin-googlemaps@2.5.0 check-types /Users/adamduren/Workspace/cordova-plugin-googlemaps
> tsc

src/browser/PluginGeocoder.js:160:18 - error TS2551: Property 'positon' does not exist on type 'GeocoderRequest'. Did you mean 'position'?

160     if (!request.positon && request.address) {
                     ~~~~~~~

  typings/index.d.ts:45:3
    45   position?: ILatLng | ILatLng[];
         ~~~~~~~~
    'position' is declared here.
```

adamduren:
Type issues would also be caught by TravisCI when merging PRs adding another level of protection.

wf9a5m75:
Moving to TypeScript enforces me to tons of work.
Because of that, I can't do that at this time, maybe in next year.

adamduren:
Thanks for taking a look, I'll keep the PR open and update the branch as I have more time. One good thing about the approach is that it's a change that can happen incrementally vs all or nothing.

wf9a5m75:
It's time to migrate to TypeScript.

PR #2520, opened 2018-12-05
Fix invalid JSON for locale "da", fix typo for locale "de"

chriz-hoepfner:
I just stumbled across two issue regarding localisation on iOS. 

The main issue looks like an unfinished refactoring when switching to JSON locales files.


wf9a5m75:
Thanks

Issue #2544, opened 2019-01-10
map.addKmlOverlay throw "self.map.addMarker(...).then is not a function"


rodneicouto:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [X] any problem or bug report

**OS:**  (check one with "x")
- [X] **Android** (it just tested here)
- [ ] **iOS**
- [ ] **Browser**

**cordova information:**  (run `$> cordova plugin list`) 
```
cordova-plugin-googlemaps 2.5.0-beta-20190102-1842 "cordova-plugin-googlemaps"
cordova-plugin-whitelist 1.3.3 "Whitelist"
```

**Current behavior:**

With some KML file, the following code launch a exception
```JAVASCRIPT
 map.addKmlOverlay({
       url: 'example.kml'
 }, function(kmlOverlay) {
     this.map.moveCamera(kmlOverlay.camera);
 })
```

The Exception
```
Uncaught (in promise) TypeError: self.map.addMarker(...).then is not a function
    at KmlLoader.js:615
    at new Promise (<anonymous>)
    at KmlLoader.parsePointTag (KmlLoader.js:591)
    at KmlLoader.parseKmlTag (KmlLoader.js:287)
    at BaseArrayClass.<anonymous> (KmlLoader.js:136)
    at BaseArrayClass.js:148
````

The KML was generated by Google MyMaps. [https://raw.githubusercontent.com/rodneicouto/cordova-googlemaps-test/master/www/example.kml](https://raw.githubusercontent.com/rodneicouto/cordova-googlemaps-test/master/www/example.kml)

**To Reproduce**
* Download repo: [https://github.com/rodneicouto/cordova-googlemaps-test](https://github.com/rodneicouto/cordova-googlemaps-test)
* Change API_KEY in both _config.xml_ and _package.json_. Look for 'CHANGE_HERE'
* run `npm install`
* run `cordova platform add android`
* run `cordova run android`

rodneicouto:
I think get the problem. In _KmlLoader.js_, line 614, is expected to _addMarker_ returns _Promise_ but the _Map.addMarker()_ returns an object of type _Marker_, not a _Promise_. 

[...]

PR #2545, opened 2019-01-11
#2544 map.addKmlOverlay throw "self.map.addMarker(...).then is not a function" 

rodneicouto:
In KmlLoader.js, line 614, is expected to addMarker returns Promise but the Map.addMarker() returns an object of type Marker, not a Promise. (issue #2544)


wf9a5m75:
This has been fixed. Thank you.

PR #2557, opened 2019-02-06
Documentation is fixed about name parameter of config.xml

kamaci:
# Pull request guide

### DIFF
--- a/src/ios/GoogleMaps/PluginEnvironment.m
+++ b/src/ios/GoogleMaps/PluginEnvironment.m
@@ -41,7 +41,7 @@ dispatch_queue_t queue;
          *
          * <?xml version='1.0' encoding='utf-8'?>
          * <widget id="(package name)" version="0.0.1" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Thank you for your pull request, but I don't think this is any problem.

kamaci:
It makes a confusion about what will be the name of the app in the hand. It doesn't have both English and Non-English app names, it will only have Non-English app name at given example.

On the other hand, English app name is not the app name, but name of the project when you open it at Xcode or Android Studio. 

I think that such a change on documentation will make it much more clear.

wf9a5m75:
Cordova CLI accepts below command.

```
$ cordova create myAppDir test.myapp.app マイプロジェクト
```
"マイプロジェクト" (=myproject) is written in Japanese, which is not English.

Project name could be non-English name. In that case, he/she makes a confusion if both the app and project name are non-English name, regarding of your description.

wf9a5m75:
So far, there is no claim about `(english app name)`. Everybody understand what that means.

wf9a5m75:
Okay, let's vote. I will leave this PR for a month.
5 people post their agree comments with clarify opinions, then I will merge this PR.

The clarify opinion means , not just "+1", but "+1 because .... that's why this PR is good for me" for example.

PR #2600, opened 2019-04-11
Update PluginPolygon.java

adappter:
Removed unnecessary code setting the visibility of the polygon hardcoded to true/false.
This is not needed, since the visibility is managed by the PolygonOptions.

### DIFF
--- a/src/android/plugin/google/maps/PluginPolygon.java
+++ b/src/android/plugin/google/maps/PluginPolygon.java
@@ -339,11 +339,6 @@ public class PluginPolygon extends MyPlugin implements MyPluginInterface  {
             public void run() {
                 // Update the polygon
                 polygon.setPoints(path);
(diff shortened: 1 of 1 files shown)


wf9a5m75:
`path.size() > 0` is necessary, because if you create a polyline without points , app would be crashed.

adappter:
Okay then I will revert it and set the polygon options visibility as parameter?

wf9a5m75:
Shit, you sent a pull request to the `master` branch.
But please create a pull request to the `multiple_maps` branch next time, as far as written in https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/master/pull_request_template.md

TdoubleG:
@webalchemist I'm 100% sure that I did it...strange. Sorry will watch better next time! Thanks for merging.

PR #2602, opened 2019-04-11
Check properties - for iOS9 compatibilty

calca:
Check if properties exist for iOS 9 compatibility


wf9a5m75:
`<engine name="apple-ios" version=">=10.0.0" />` in the plugin.xml.

It means ios9 is not supported.

Issue #2627, opened 2019-05-30
KmlLoader unable to load KML with <color>00000000</color>


carlchan1994:
Some KML file have a # before the color, and some do not:

`<color>#00000000</color>`
`<color>00000000</color>`

The colorStr(00000000)  seems to be read as a number instead of string and an error of "colorStr.replace is not a function" happens when loading KML.

=======================================================

> //-------------------------------
> // KML color (AABBGGRR) to RGBA
> //-------------------------------
> function kmlColorToRGBA(colorStr) {
>   var rgba = [];
>   colorStr = colorStr.replace('#', '');
>   for (var i = 6; i >= 0; i -= 2) {
>     rgba.push(parseInt(colorStr.substring(i, i + 2), 16));
>   }
>   return rgba;
> }

========================================================

> ERROR Error: Uncaught (in promise): TypeError: colorStr.replace is not a function
> TypeError: colorStr.replace is not a function
>     at kmlColorToRGBA (KmlLoader.js:1066)
>     at KmlLoader.js:743
>     at Array.forEach (<anonymous>)
>     at KmlLoader.js:740
>     at Array.forEach (<anonymous>)
>     at KmlLoader.parseLineStringTag (KmlLoader.js:736)
>     at KmlLoader.parseKmlTag (KmlLoader.js:302)
>     at BaseArrayClass.<anonymous> (KmlLoader.js:135)
>     at BaseArrayClass.js:147
>     at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:391)
>     at resolvePromise (zone.js:831)
>     at zone.js:896
>     at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:423)
>     at Object.onInvokeTask (core.js:17290)
>     at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:422)
>     at Zone.push../node_modules/zone.js/dist/zone.js.Zone.runTask (zone.js:195)
>     at drainMicroTaskQueue (zone.js:601)
>     at ZoneTask.push../node_modules/zone.js/dist/zone.js.ZoneTask.invokeTask [as invoke] (zone.js:502)
>     at invokeTask (zone.js:1744)
>     at Worker.globalZoneAwareCallback (zone.js:1770)
> defaultErrorLogger @ core.js:15724

[...]

PR #2628, opened 2019-05-30
Fix KmlLoader unable to read pure digit color <color>00000000</color> #2627

carlchan1994:
Fix KmlLoader unable to read pure digit color <color>00000000</color> #2627


wf9a5m75:
This is already fixed. Thank you for pointing out

PR #2674, opened 2019-08-28
Add Heatmap support in Android to plugin

Lock-n:
I added a new class "Heatmap" that allows drawing of heatmaps using [Google's Heatmap Utility](https://developers.google.com/maps/documentation/android-sdk/utility/heatmap) on Android (no support to iOS).

The classes `Heatmap.js`, `PluginHeatmap.js`, `PluginHeatmap.java` were all implemented and integrated to the main class `Map.js`.

The code works, although probably some functions could've been better implemented (I did this in a hurry).

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -40,6 +40,9 @@
   <js-module name="Circle" src="www/Circle.js">
     <runs/>
   </js-module>
+  <js-module name="Heatmap" src="www/Heatmap.js">
+    <runs/>
+  </js-module>
   <js-module name="TileOverlay" src="www/TileOverlay.js">
     <runs/>
   </js-module>
@@ -124,7 +127,7 @@
 
   <engines>
     <engine name="cordova" version=">=7.0.0" />
-    <engine name="cordova-ios" version=">=4.5.0" />
+    <engine name="cordova-ios" version=">=5.0.0" />
     <engine name="apple-xcode" version=">=9.0.0" />
     <engine name="apple-ios" version=">=10.0.0" />
   </engines>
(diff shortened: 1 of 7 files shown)


Lock-n:
Maybe you could add this feature to the plugin and disclose that this specific way to draw heatmaps uses a Google library and possibly could cause unexpected things. I imagine that there is other people like me that will need to draw heatmaps with this plugin, and for us this feature is very useful.

What do you think?

I also improved my code and have an updated version of it.

wf9a5m75:
One day, some one reports an unknown issue. This plugin provides Android, iOS, and JavaScript. These implementations are different, which are used other people code.
So who will need to inspect the code? Are you in charge of that?

Debugging of other people's code is heavy weight task.
Can you in charge of the role forever?

What do you think?

Lock-n:
You could have a disclaimer telling that this specific feature may have bugs and you won't correct them. There's really no way of this piece of code generating bugs unless someone calls it. So, you could add it as an additional, with no guarantee of stability, feature. Until you finish your implementation of a heatmap, this can be practical for a lot of people.

But, of course, it's your decision whether to add this or not. I get that you're worried that the code might bring instability to the plugin. Maybe this could be in another branch, or in folder in case someone wants to use it? Something like that?

wf9a5m75:
Why don't you implement on your repo and provide it by yourself?
You don't need to send a pull request mandatory.

Under the Apache license 2, you can re-distribute your own version.

I only accept pull requests either fixing bug, and improve the code without dependency and implement the same code of three platforms with the same logic.

adamduren:
@wf9a5m75 This would be a more broad discussion outside of this issue but it would be awesome if we could develop a way to extend `cordova-plugin-googlemaps` via a well defined extension api. 

If this was achievable, `cordova-plugin-googlemaps` could focus on core functionality and others could contribute plugins to extend the functionality.

wf9a5m75:
Agree. I considered couple of times about that. But it's not a good time for me now.

PR #2681, opened 2019-09-16
Fix null exception

8bhsolutions:
When calling setDiv(null) in the js page, however, map resize event is also happening. This causes null exception.

**Scenario**.
_App setup_
Root page has a map view. When pushing a child view (which also has another map view), I call map.setDiv(null) in the root page willLeave function.

_Steps_
I load my app through deeplinks to view child map. It start load of root page, then automatically push child page. This scenario occasionally encounter this null exception

# Pull request guide


wf9a5m75:
Thank you. I appreciate for your help!

PR #2691, opened 2019-10-07
fixed flickering and wrong rendering of some dom elements

tomaz-dev:
Browser was trying to set index.html itself as background image. That invoked a warning: Resource interpreted as Image but transferred with MIME type text/html: "file:///android_asset/www/index.html". For the time of downloading index.html there was a flicker and some elements was not renered properly.


wf9a5m75:
I see. Thanks!

PR #2694, opened 2019-10-23
Use an actual semaphore to manage the cleanup and always release the lock

JordyLangen:
In the `PluginMarker.java` the semaphore was always waiting until it expired it's wait time. After investigating it seemed that `keySet` was empty in our case, and the cleanDone was not set to true, nor was the lock released. After fixing this I get an exception that told me the semaphore was not locked on that thread, so the implementation now uses an actual semaphore.


wf9a5m75:
Thanks for your contribution!

Issue #2717, opened 2019-12-12
[Android] Clear map block map


goinnn:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [x] any problem or bug report

**OS:**  (check one with "x")
- [x] **Android**
- [ ] **iOS**
- [ ] **Browser**

**cordova information:**  (run `$> cordova plugin list`) 

```
cordova-android-firebase-gradle-release 3.0.0 "cordova-android-firebase-gradle-release"
cordova-android-play-services-gradle-release 3.0.0 "cordova-android-play-services-gradle-release"
cordova-android-support-gradle-release 3.0.1 "cordova-android-support-gradle-release"
cordova-open-native-settings 1.5.2 "Native settings"
cordova-plugin-actionsheet 2.3.3 "ActionSheet"
cordova-plugin-advanced-http 2.1.1 "Advanced HTTP plugin"
cordova-plugin-camera 4.0.3 "Camera"
cordova-plugin-core-android-extensions 2.0.0 "CoreAndroidExtensions"
cordova-plugin-device 2.0.3 "Device"
cordova-plugin-dialogs 2.0.2 "Notification"
cordova-plugin-file 6.0.1 "File"
cordova-plugin-file-transfer 1.7.1 "File Transfer"
cordova-plugin-filepath 1.5.5 "cordova-plugin-filepath"
cordova-plugin-geolocation 4.0.1 "Geolocation"
cordova-plugin-googlemaps 2.6.2 "cordova-plugin-googlemaps"
cordova-plugin-headercolor 1.0 "HeaderColor"
cordova-plugin-inappbrowser 3.0.0 "InAppBrowser"
cordova-plugin-ionic-webview 1.1.16 "cordova-plugin-ionic-webview"
cordova-plugin-localization-strings 3.1.0 "Localization"
cordova-plugin-nativeaudio 3.0.9 "Cordova Native Audio"
cordova-plugin-network-information 2.0.3-dev "Network Information"
cordova-plugin-screen-orientation 3.0.2 "Screen Orientation"
cordova-plugin-splashscreen 5.0.2 "Splashscreen"
cordova-plugin-statusbar 2.4.3 "StatusBar"
cordova-plugin-transparent-status-bar 1.0.1 "Transparent status bar"
cordova-plugin-whitelist 1.3.4 "Whitelist"
[...]

PR #2726, opened 2019-12-16
Fixing clear map blocked 1 second

goinnn:
Fixing #2717 

This fixed a bug: 759c725
This is only an opinion: a7a23fb. We prefer 5 (like in 2.1.1 version) than 1000


wf9a5m75:
Another PR should solve this problem. But thanks anyway.

PR #2728, opened 2019-12-16
Fix two js error when change very quickly between two pages with map

goinnn:
Fixing #2718

**Incomplete**: This PR only fixes polylines & markers


wf9a5m75:
Thanks for your contribution!

Issue #2748, opened 2020-03-08
bug: NullPointer Exception while „stopAnimation“


HansKrywaa:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [X] any problem or bug report

**OS:**  (check one with "x")
- [X] **Android**
- [ ] **iOS**
- [ ] **Browser**

**Current behavior:**

We just replaced our long live native App with a hybrid one including this Plugin. In the Crash Reports we have some things now, for Example:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void com.google.android.gms.maps.GoogleMap.stopAnimation()' on a null object reference

plugin.google.maps.PluginMap$13.run PluginMap.java:1767
android.os.Handler.handleCallback Handler.java:789
android.os.Handler.dispatchMessage Handler.java:98
android.os.Looper.loop Looper.java:164
android.app.ActivityThread.main ActivityThread.java:6944
java.lang.reflect.Method.invoke Method.java
com.android.internal.os.Zygote$MethodAndArgsCaller.run Zygote.java:327
com.android.internal.os.ZygoteInit.main ZygoteInit.java:1374
```

**Expected behavior:**

I already create a PR to fix this: #2749 - Please merge this

wf9a5m75:
The PR has been merged on the multiple_maps branch.

HansKrywaa:
Thank you! 💪🏼

PR #2749, opened 2020-03-08
fixed NullPointerException while „stopAnimation“

HansKrywaa:
Related Issue: #2748

This should fix the NullPointerException -> App Crash described in the relative issue.


HansKrywaa:
Please merge this 🙈 We have 5 Crash Reports about that now

Issue #2746, opened 2020-03-05
Plugin is not compatible with iOS from April 2020


chris087:
This plugin uses the uiwebview, which is not supported by Apple from April 2020.
@see: https://ionicframework.com/blog/understanding-itms-90809-uiwebview-api-deprecation/

The plugin must be updated to use WKWebView.

wf9a5m75:
This plugin also supports WKWebView on ionic.

Mathieu-Castets:
Yep. But action are needed by the end of 2020 for existing apps and by April 2020 for new apps using this plugin.

Every plugin should understand the new config.xml preference tag: `<preference name="WKWebViewOnly" value="true" />` introduced with `cordova-ios@5.0.1`.

**References to UIWebView must be removed from the source.**

See also other plugins such as InAppBrowser which are taking actions: https://github.com/apache/cordova-plugin-inappbrowser/pull/584

wf9a5m75:
Ah, I see. Please send a pull request.

HansKrywaa:
Anyone working on a PR? Otherwise i will, as the Changes seems not to be that hard?

Mathieu-Castets:
> Anyone working on a PR? 

I don't have enough Objective-C background to do so but I could try based on what have been done on InAppBrowser plugin. The main strategy seems to add a condition on every places containing the UiWebView keyword if WkWebViewOnly preference is present.

wf9a5m75:
The PR has been merged on the multiple_maps branch.

wf9a5m75:
The cordova-plugin-googlemaps version 2.7.0 has been released.

HansKrywaa:
@wf9a5m75 Is there a Changelog for 2.7.0? I like installing a stable Version more than a Branch, so is 2.7.0 the old multiple_maps?

battika:
Thanks Masashi.

wf9a5m75:
Sorry for late action.

PR #2752, opened 2020-03-09
feat: Remove UIWebView Code completely 

HansKrywaa:
# Pull request guide

**Related Issue**: #2746 

I'm not that much into native iOS Development, but i tried my best. I used the [PR](https://github.com/apache/cordova-plugin-inappbrowser/pull/584), mentioned in the related Issue, as template.

- I searched for `UiWebView` in Code and found Usages in:
  - PluginGroundOverlay.m
  - PluginKmlOverlay.m
  - PluginMarker.m
  - PluginTileOverlay.m
  - PluginUtil.h & PluginUtil.m
- I added the `preprocessing #if directives`
- Because the WkWebView Part can be executed from two Places (WK_WEB_VIEW_ONLY is set or `![clsName isEqualToString:@"UIWebView"]`) i created a Methode for this (exclude `PluginTileOverlay.m` as it is just one line)
- I changed the Header File (add the new Methode or edit existing directives (`PluginUtil.h`)
- `npm run test` runs successfully

Seems to work for me - Please give it a Look 🍺

### DIFF
--- a/src/ios/GoogleMaps/PluginGroundOverlay.h
+++ b/src/ios/GoogleMaps/PluginGroundOverlay.h
@@ -25,7 +25,6 @@
 - (void)setBearing:(CDVInvokedUrlCommand*)command;
 - (void)setZIndex:(CDVInvokedUrlCommand *)command;
 - (void)_setImage:(GMSGroundOverlay *)layer urlStr:(NSString *)urlStr completionHandler:(void (^)(BOOL succeeded))completionHandler;
-- (void)_WKsetImage:(GMSGroundOverlay *)groundOverlay webview:(id)webview urlStr:(NSString *)urlStr completionHandler:(void (^)(BOOL succeeded))completionHandler;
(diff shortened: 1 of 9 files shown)


wf9a5m75:
Thank you for your PR. Let's remove all the code for UIWebView at this time. This action drops some of older Cordova versions, but Apple won't accept the Apps which includes the code of UIWebView anyway. So I think we don't need to take care for UIWebView anymore eventually.

HansKrywaa:
@wf9a5m75 What about the `webViewDidFinishLoad()` Methode in `PluginUtil.m`. Currently this only works with UiWebView, should it be completely removed or changed to WkWebView?

HansKrywaa:
Ok, i removed the `preprocessing  directives` and the existing If Statements and just keep the WkWebView Code.

wf9a5m75:
Sorry for late response. I will check this later today

Issue #2753, opened 2020-03-10
bug: App Crash - java.util.ConcurrentModificationException


HansKrywaa:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [X] any problem or bug report

**OS:**  (check one with "x")
- [X] **Android**
- [ ] **iOS**
- [ ] **Browser**

**Current behavior:**

We had a new Crash Report in our App using this Plugin:
```
java.util.ConcurrentModificationException: java.util.ConcurrentModificationException

java.util.LinkedHashMap$LinkedHashIterator.nextNode LinkedHashMap.java:757
java.util.LinkedHashMap$LinkedValueIterator.next LinkedHashMap.java:785
plugin.google.maps.CordovaGoogleMaps.onResume CordovaGoogleMaps.java:527
org.apache.cordova.PluginManager.onResume PluginManager.java:264
org.apache.cordova.CordovaWebViewImpl.handleResume CordovaWebViewImpl.java:452
org.apache.cordova.CordovaActivity.onResume CordovaActivity.java:277
android.app.Instrumentation.callActivityOnResume Instrumentation.java:1454
android.app.Activity.performResume Activity.java:8129
android.app.ActivityThread.performResumeActivity ActivityThread.java:4400
android.app.ActivityThread.handleResumeActivity ActivityThread.java:4442
android.app.servertransaction.ResumeActivityItem.execute ResumeActivityItem.java:52
android.app.servertransaction.TransactionExecutor.executeLifecycleState TransactionExecutor.java:176
android.app.servertransaction.TransactionExecutor.execute TransactionExecutor.java:97
android.app.ActivityThread$H.handleMessage ActivityThread.java:2146
android.os.Handler.dispatchMessage Handler.java:107
android.os.Looper.loop Looper.java:237
android.app.ActivityThread.main ActivityThread.java:7777
java.lang.reflect.Method.invoke Method.java
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run RuntimeInit.java:493
com.android.internal.os.ZygoteInit.main ZygoteInit.java:1047
```

**Expected behavior:**

I already create a PR for this. Please check this out: #2754

wf9a5m75:
The PR has been merged on the multiple_maps branch.

HansKrywaa:
Thank you 💪🏼

PR #2754, opened 2020-03-10
Fix App Crash on Resume with java.util.ConcurrentModificationException

HansKrywaa:
#Pull Request

**Related Issue**: #2753

Please give it a Look and merge 😊

### DIFF
--- a/src/android/plugin/google/maps/CordovaGoogleMaps.java
+++ b/src/android/plugin/google/maps/CordovaGoogleMaps.java
@@ -40,7 +40,6 @@ import org.json.JSONObject;
 
 import java.lang.reflect.Field;
 import java.util.Collection;
-import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Please write the code for Java 1.6 to older environment (unfortunately).

HansKrywaa:
I'm not sure, but you mean use
```
for(PluginEntry p: pluginsToResume) {}
```

instead of the `forEach` Loop?

wf9a5m75:
Yes. The most `Cordova` users are familiar with only JavaScript, not familiar with Java or Objective-C.
And many users don't update the JDK, or face to trouble to upgrade the JDK version.
That's why they face compile error, then they claim to this repository (because those people can not understand the reason).

I faced the similar cases many times, and they gets me riled up.
In order to avoid this, I write the all code of this plugin with `JDK 1.6` (Java 7).
Thank you for your understanding and cooperation.

HansKrywaa:
Okay, i understand. Thank you for your Explanation. When i search through Google after the Error, i found only Results while removing or adding items while iterating, but that not the Case here? Regarding to [this](https://stackoverflow.com/questions/3184883/concurrentmodificationexception-for-arraylist) the Usage of a Iterator should help, am i correct?

- PR is updated

Issue #2762, opened 2020-04-09
UiWebView references present in v2.7.0


dondos:
Related issue: #2756

Apple still complains about uiwebview references in my app, which was build using version 2.7.0 of this plugin. The only file referencing that was PluginMarker.m:
https://github.com/mapsplugin/cordova-plugin-googlemaps/blob/1e27920751a74fb015dc81618e78e369f3d5b1fa/src/ios/GoogleMaps/PluginMarker.m#L1146-L1147
Shouldn't that reference also be removed?

wf9a5m75:
Got it. Thank you for your report.

wf9a5m75:
Please reinstall v2.7.1

PR #2771, opened 2020-05-05
Fix: UiWebView references present in v2.5.1 #2762

ronaldozanoni:
This PR is only to apply the commit to remove the UIWebView code's references in the version 2.5.1.
In my case, I cannot use any version above 2.5.1 (2.5.2, 2.5.3, 2.7.x etc) because that versions cause issues when clicking on bottom buttons in the iPhone X or above.

I haven't investigated why this issue occurs in that versions yet, so for now, I would like to keep using the **v2.5.1**.

See the gif below showing the issue with the iPhone X.

iPhoneX-bottom-click-issue

Could you please merge that PR?

Thanks.


wf9a5m75:
No longer support for v2.5.1

wf9a5m75:
> because that versions cause issues when clicking on bottom buttons in the iPhone X or above.

If you have a click problem, please create another issue, and provide an example project.

PR #2777, opened 2020-05-16
[Android] Fix null exception when clicking on polyline when there are multiple polyline on the map

8bhsolutions:
The touchPoint is overwritten by the result of the polyline with highest zIndex - which could be null

# Pull request guide

### DIFF
--- a/src/android/plugin/google/maps/PluginMap.java
+++ b/src/android/plugin/google/maps/PluginMap.java
@@ -3031,6 +3031,8 @@ public class PluginMap extends MyPlugin implements OnMarkerClickListener,
             Object overlay;
             String key;
 
+            double currDistance = -1;
+
             while(iterator.hasNext()) {
               entry = iterator.next();
               key = entry.getKey();
@@ -3041,12 +3043,10 @@ public class PluginMap extends MyPlugin implements OnMarkerClickListener,
                 if (polyline == null) {
                   continue;
                 }
-                zIndex = polyline.getZIndex();
-                if (zIndex < maxZIndex) {
-                  continue;
-                }
 
+                zIndex = polyline.getZIndex();
                 points = polyline.getPoints();
+                LatLng touchPoint_OnPolyline;
 
                 if (polyline.isGeodesic()) {
                   hitArea.x = (int)(polyline.getWidth() * density);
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Thank you for sending your PR. Please give me the test code

8bhsolutions:
@wf9a5m75 
I didn't see any unit/automated test framework in the plugin

Can you explain what test code you are looking for?

wf9a5m75:
Please create the code to reproduce the `null pointer` error test case.
I mean how can I confirm your `null exception`

8bhsolutions:
OK.
The null exception occurred with production data, so I can't share the user data.

I'll try reproduce it, with a different set of data points, and will share code once I reproduce it.

**SIDE NOTE**
While this fix, resolve my instance of the problem. When I was debugging this, I felt that the problem could be in **isPointOnTheGeodesicLine**, as it returned null.
The "unfix" code could work if isPointOnTheGeodesicLine does not return null.

I tried following the code, but couldn't quite follow the geometric math. So I can't say if the algorithm/logic in isPointOnTheGeodesicLine is correct in returning null for the specific data.

wf9a5m75:
This is solved in https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/2786

8bhsolutions:
Thanks @wf9a5m75 .

Haven't had enough time to produce a dataset that would reproduce the null error in the code.

#2786 is a good description of the issue of why when null is returned by either isPointOnTheGeodesicLine or isPointOnTheLine. It is then return to the javascript code.


The fix you have implemented in multiple_maps branch is basically the same as this pull request. It uses a separate variable for polyline touch instead of assigning directly to touchPoint variable

However, in addition to that, this PR also keeps track of the closest touchPoint_onPolyline to the user touchPoint.

This is needed because there could be mutliple polylines near the user touch point. The code will loop through all the polylines. 
Without keeping track of the closet point, the code just returns the last polyline in the loop where the isPointOnPolyline is true, which may not necessarily be the closet point to the user touch point.

wf9a5m75:
Good point, but I will change the implementation of this in more effort way.

8bhsolutions:
Thanks @wf9a5m75

Issue #1183, opened 2016-11-12
Disable 3D buildings


gigiojr:
I found a Android [documentation](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#setBuildingsEnabled%28boolean%29) about disable 3D buildings on Google Maps API using the method `setBuildingsEnabled(false)`, but I don't found nothing in plugin documentation and I need that this works for Android, IOs and Windows Phone.

Is there some solution to help me?

wf9a5m75:
This plugin does not work on Windows phone (no plan to implement).
It seems I forgot to implement the  setBuildingsEnabled(false) method.

wf9a5m75:
I will implement this for both v1.4 and v2.0 in couple of days.

wf9a5m75:
Ok, you can disable the building like this:

```js
var map = plugin.google.maps.Map.getMap(div, {
   'preferences': {
     'building': false
   }
});
```

Please reinstall the plugin from the master branch.
```
$> cordova plugin rm cordova-plugin-googlemaps
$> cordova plugin rm cordova-plugin-googlemaps-sdk

$> cordova plugin add https://github.com/mapsplugin/cordova-plugin-googlemaps --variable ...
```

moriczgergo:
For some reason the aren't disabled, even when I'm using {preferences: {building: false}}.

adamduren:
@moriczgergo Can you share your project or a code sample to reproduce the issue?

adamduren:
Also what platform(s) and version of the plugin?

alexp25:
It seems that using {preferences: {building: false}} does not work. Maybe the setBuildingsEnabled method should be implemented.

PR #2779, opened 2020-05-20
[ANDROID] [Feature] - Ability to change clickability of POI

8bhsolutions:
This PR introduce the ability to change whether POI is clickable.

If the developer does not want interactions with POI, currently, the plugin can only hide the POI by setting map styles.
This give the developer more flexibly on how to handle POI on their map.

**Use case**
Sometime tapping on Polyline does not register. This is due to the tap position being near a POI. The POI registers the clicking event instead of the Polyline.

# Pull request guide

----

@wf9a5m75
**master** branch has changes that is ahead of **multiple_maps** branch. Creating a PR into **multiple_maps** will also bring these changes.

### DIFF
--- a/src/android/plugin/google/maps/PluginMap.java
+++ b/src/android/plugin/google/maps/PluginMap.java
@@ -3179,5 +3179,20 @@ public class PluginMap extends MyPlugin implements OnMarkerClickListener,
     }
   }
 
+  public void setClickablePOI(JSONArray args, final CallbackContext callbackContext)  throws JSONException {
+    final boolean clickable = args.getBoolean(0);
+    activity.runOnUiThread(new Runnable() {
+      @Override
+      public void run() {
+        if (clickable) {
+          map.setOnPoiClickListener(PluginMap.this);
+        } else {
+          map.setOnPoiClickListener(null);
+        }
 
+        callbackContext.success();
+      }
+    });
+  }
+  
 }
--- a/www/Map.js
+++ b/www/Map.js
@@ -1586,4 +1586,15 @@ Map.prototype._onCameraEvent = function(eventName, cameraPosition) {
   }
 };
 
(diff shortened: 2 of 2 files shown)


wf9a5m75:
```js
map.on(event.POI_CLICK, function(placeId, name, latLng) {
    
});
```

wf9a5m75:
I implemented `mapOptions.preferences.building` flag before, but somehow I dropped the flag.

https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/1183

wf9a5m75:
Since Google Maps JS v3 does not have the name of method `setClickablePOI`, I don't accept this style way.
The goal of this plugin is to reproduce the Google Maps JS v3 as much as possible.

I'm implementing `mapOptions.preferences.clickableIcons` option more better way. I will let you know the code is ready.

wf9a5m75:
`mapOptions.preferences.clickableIcons = true/false` is now available on the multiple_maps branch code.

8bhsolutions:
No worries. Noted about conforming to Google Maps JS API.

Regarding your code change. I don't believe it will solve the issue.

The reason the Polyline click is not registering is because the POI event listener is registered.
`map.setOnPoiClickListener(PluginMap.this)`
Because of this, **onPoiClick** will consume the click event instead of **onMapClick**.

The Java code has to set the listener to null so that the click event can be consumed by onMapClick instead.
`map.setOnPoiClickListener(null)`

8bhsolutions:
Thanks mate

wf9a5m75:
```java
  public void onPoiClick(PointOfInterest pointOfInterest) {
    if (!this.clickableIcons) {
      this.onMapClick(pointOfInterest.latLng);
      return;
    }
```

8bhsolutions:
Yep. Noticed that after my previous comment :)

PR #2811, opened 2020-07-15
Add "number" parameter to MarkerCluster

rcpinheiro:
When you add markers to  MarkerCluster add a optional "number" parameter to the MarkerOptions (default: number=1). That way, MarkerCluster instead of showing the count of Markers, will show the sum of the "number" field in the Markers.
Very useful when you try to add lots and lots of identifical markes. For instance, if you have 100 markers in the same location, you only have to add one marker with "number=100" instead of adding 100 markers.

# Pull request guide

### DIFF
--- a/www/MarkerCluster.js
+++ b/www/MarkerCluster.js
@@ -1003,10 +1003,15 @@ Object.defineProperty(MarkerCluster.prototype, '_redraw', {
         }
 
         unionedMarkers.forEach(function (cluster) {
-
+          var number=0;
+          for (var f=0;f<cluster._markerArray.length;f++) {
+            if (typeof cluster._markerArray[f].get('number')!=undefined) {
+              number+=parseInt(cluster._markerArray[f].get('number'));
+            }
+          }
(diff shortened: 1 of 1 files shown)


wf9a5m75:
Could you show me the screenshot? (I'm outside now)

wf9a5m75:
Ah, I got your idea.
You want to show the `value of markers` instead of `number of markers`.

rcpinheiro:
Right.
Three 3 markers with "number" equal to 48, 11 and 2:
image

When you zoom out:
image

rcpinheiro:
On second thought, the name "number" is not a good one. It's best to call it "value" or "weight".

wf9a5m75:
@rcpinheiro 
Thank you for your PR. Your idea is good, but creating a special meaning property for particular is not good.
Because some people might want to more complex calculation, such as `return value * priority;`

I also asked someone wanted to customize the strings of the number, such as `2532 -> 2.5k`

So, I will create a custom callback instead of particular property.

PR #2821, opened 2020-08-05
Fixed bug `Cannot redefine property: waypoints`

timkovik:
This bug was excepted while try execute addDirectionsRendererSync() method

### DIFF
--- a/www/DirectionsRenderer.js
+++ b/www/DirectionsRenderer.js
@@ -52,10 +52,7 @@ var DirectionsRenderer = function(map, exec, options) {
     value: new BaseArrayClass(),
     writable: false
   });
-  Object.defineProperty(self, 'waypoints', {
-    value: new BaseArrayClass(),
-    writable: false
-  });
+
 
   Object.defineProperty(self, 'pathList', {
     value: new BaseArrayClass(),


wf9a5m75:
When do you need to redefine waypoints?

wf9a5m75:
The `waypoints` property is designed as `non-redefine` property.
You can manipulate it through the methods.
https://github.com/mapsplugin/cordova-plugin-googlemaps-doc/blob/master/v2.0.0/class/BaseArrayClass/README.md

By the way, Directions class is not ready to use in public. It means you can not use it in Android and iOS yet.

timkovik:
> When do you need to redefine waypoints?

https://github.com/mapsplugin/ionic-googlemaps-quickdemo-v5/blob/b8efddb1d8238603168e47789ae713b4924ac230/src/app/directions/directions.page.ts#L75 in this example

timkovik:
> Кстати, класс Directions не готов к публичному использованию. Это означает, что вы пока не можете использовать его в Android и iOS.

What it mean? Is it mean that it will not work on production-build app?

And this words sound like `you cant draw routes on your native map with this plugin` 😟

wf9a5m75:
```typescript
if (!this.renderer) {
        // Only first time
      } else {
        // Second time or later
      }
```

wf9a5m75:
> And this words sound like you cant draw routes on your native map with this plugin

I said so.

PakMan6:
Hi guys, Great job so far. Looking forward to android and ios Directions functionality.  I want to stop using maps javascript Api.. but only when the Directions service works well for native devices. Any ideas on when Directions class will be ready for android and ios?
I noticed a few bugs in browser view as well .. addDirectionsRendererSync options (DirectionsRendererOptions) >> suppressMarker, polylineOptions and a few others don't seem to worrk currently.

sachi097:
Hi @timkovik 
Does this fix work on production build app?

Were you able to get directions in production build app after this fix?

Issue #2845, opened 2020-12-05
Polyline_click not fired or fired in a wrong place with a lot of different drawed polylines


boblepepeur:
**I'm submitting a ...**  (check one with "x")
- [ ] question
- [x] any problem or bug report

**OS:**  (check one with "x")
- [ ] **Android**
- [x] **iOS**
- [ ] **Browser**

**cordova information:**  (run `$> cordova plugin list`) 

```
com.googlemaps.ios 3.9.0 "Google Maps SDK for iOS"
cordova-pdf-generator 2.1.1 "PDFGenerator"
cordova-plugin-app-version 0.1.9 "AppVersion"
cordova-plugin-background-geolocation 3.1.0 "CDVBackgroundGeolocation"
cordova-plugin-bluetoothle 6.0.2 "Bluetooth LE"
cordova-plugin-device 2.0.2 "Device"
cordova-plugin-file-opener2 3.0.5 "File Opener2"
cordova-plugin-geolocation 4.0.2 "Geolocation"
cordova-plugin-globalization 1.11.0 "Globalization"
cordova-plugin-googlemaps 2.7.1 "cordova-plugin-googlemaps"
cordova-plugin-ionic-keyboard 2.2.0 "cordova-plugin-ionic-keyboard"
cordova-plugin-ionic-webview 4.2.1 "cordova-plugin-ionic-webview"
cordova-plugin-network-information 2.0.2 "Network Information"
cordova-plugin-splashscreen 5.0.2 "Splashscreen"
cordova-plugin-statusbar 2.4.2 "StatusBar"
cordova-plugin-whitelist 1.3.3 "Whitelist"
cordova-sqlite-storage 5.1.0 "Cordova sqlite storage plugin - cordova-sqlite-storage plugin version"
```

**If you use `@ionic-native/google-maps`, please tell the package.json (only `@ionic-native/core` and `@ionic-native/google-maps` are fine mostly)**
```
@ionic-native/core : 
{
  "_from": "@ionic-native/core@^5.0.0",
  "_id": "@ionic-native/core@5.28.0",
  "_inBundle": false,
  "_integrity": "sha512-YmR70U+bIk1e4V36bke8HCol2XwuplAr4OTok4Jdx6MaAfJ1xR1AWt2PiQOIgM8bAq8qkVlbgsYi2rEOzKVc7A==",
  "_location": "/@ionic-native/core",
  "_phantomChildren": {},
  "_requested": {
[...]

PR #2847, opened 2020-12-11
Fix tap on polyline

boblepepeur:
According to my issue -> https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/2845

I use the lib & possibility of google map sdk native plugin on ios.
When you have a lot of polylines in your map, the previous fonctionnality of this cordova plugin to try to calculate if click is on polyline don't work, so i use the native default functionnality with didtabOverlay event.
The only default of my solution is that i cannot find the position clicked by user, the google map sdk plugin don't give the position clicked. So i return the center of the clicked polyline.

### DIFF
--- a/src/ios/GoogleMaps/PluginPolyline.m
+++ b/src/ios/GoogleMaps/PluginPolyline.m
@@ -102,12 +102,12 @@
 
       // Since this plugin provide own click detection,
       // disable default clickable feature.
-      polyline.tappable = NO;
+      polyline.tappable = [[json valueForKey:@"clickable"] boolValue];
 
       NSString *id = [NSString stringWithFormat:@"polyline_%@", idBase];
       [self.mapCtrl.objects setObject:polyline forKey: id];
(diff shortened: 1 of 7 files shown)


src/ios/GoogleMaps/PluginPolyline.m
```diff
@@ -102,12 +102,12 @@ -(void)create:(CDVInvokedUrlCommand *)command
 
       // Since this plugin provide own click detection,
       // disable default clickable feature.
-      polyline.tappable = NO;
+      polyline.tappable = [[json valueForKey:@"clickable"] boolValue];
```
wf9a5m75, line 105: This is wrong. If you set `polyline.tappable = YES`, this plugin can not get clicked latlng.

boblepepeur: I agree with you, but the code to try to find if the click is on the polyline don't work see my git project example in #2845

